Issue #373, opened 2015-01-09
Problems configuring CMake GUI


thomthom:
Disclaimer - I've not used CMake before, with all possibility the issue is behind the keyboard.

I'm getting errors when I try to load up the project in CMake GUI. I want to build OpenSubDiv without all the optional parts - so simply just the core. But I'm getting errors when I load up the CMake configuration:

![image](https://cloud.githubusercontent.com/assets/192418/5688582/3d1c60e6-9859-11e4-9448-7bbbb27708b2.png)

```
The C compiler identification is MSVC 18.0.31101.0
The CXX compiler identification is MSVC 18.0.31101.0
Check for working C compiler using: Visual Studio 12 2013
Check for working C compiler using: Visual Studio 12 2013 -- works
Detecting C compiler ABI info
Detecting C compiler ABI info - done
Check for working CXX compiler using: Visual Studio 12 2013
Check for working CXX compiler using: Visual Studio 12 2013 -- works
Detecting CXX compiler ABI info
Detecting CXX compiler ABI info - done
Compiling OpenSubdiv version v3_0_0_beta
Using cmake version 3.0.1
Try OpenMP C flag = [/openmp]
Performing Test OpenMP_FLAG_DETECTED
Performing Test OpenMP_FLAG_DETECTED - Success
Try OpenMP CXX flag = [/openmp]
Performing Test OpenMP_FLAG_DETECTED
Performing Test OpenMP_FLAG_DETECTED - Success
Found OpenMP: /openmp  
Could NOT find TBB (missing:  TBB_INCLUDE_DIR TBB_LIBRARIES) (Required is at least version "4.0")
Found OpenGL: opengl32  
Could NOT find CLEW (missing:  CLEW_INCLUDE_DIR CLEW_LIBRARY) 
Could NOT find OpenCL (missing:  OPENCL_LIBRARIES OPENCL_INCLUDE_DIRS) (Required is at least version "1.1")
CUDA_TOOLKIT_ROOT_DIR not found or specified
Could NOT find CUDA (missing:  CUDA_TOOLKIT_ROOT_DIR CUDA_NVCC_EXECUTABLE CUDA_INCLUDE_DIRS CUDA_CUDART_LIBRARY) (Required is at least version "4.0")
Could NOT find GLFW (missing:  GLFW_INCLUDE_DIR GLFW_LIBRARIES) (Required is at least version "3.0.0")
[...]

Issue #358, opened 2014-10-15
glViewer 'Failed to open window'


deepak1556:
OS - archlinux 64-bit
glfw-3.0.4-2 , glew-1.11.0-1
opensubdiv compiled successfully

```
bin git:master  ❯ sudo lshw -short | grep -i --color display
/0/100/1/0             display        GK107M [GeForce GT 755M]
/0/100/2               display        4th Gen Core Processor Integrated Graphics Controller

bin git:master  ❯ ldd glViewer       
    linux-vdso.so.1 (0x00007fffb7ffc000)
    libpng16.so.16 => /usr/lib64/lib64/libpng16.so.16 (0x00007f23b954b000)
    libz.so.1 => /usr/lib64/lib64/libz.so.1 (0x00007f23b9335000)
    libosdCPU.so.3.0.0.alpha => /home/robo/github/opensubdiv/build/lib/libosdCPU.so.3.0.0.alpha (0x00007f23b9043000)
    libosdGPU.so.3.0.0.alpha => /home/robo/github/opensubdiv/build/lib/libosdGPU.so.3.0.0.alpha (0x00007f23b8dff000)
    libGLU.so.1 => /usr/lib64/lib64/libGLU.so.1 (0x00007f23b8b7e000)
    libGL.so.1 => /usr/lib64/lib64/libGL.so.1 (0x00007f23b890e000)
    libSM.so.6 => /usr/lib64/lib64/libSM.so.6 (0x00007f23b8706000)
    libICE.so.6 => /usr/lib64/lib64/libICE.so.6 (0x00007f23b84e9000)
    libX11.so.6 => /usr/lib64/lib64/libX11.so.6 (0x00007f23b81a7000)
    libXext.so.6 => /usr/lib64/lib64/libXext.so.6 (0x00007f23b7f95000)
    libglfw.so.3 => /usr/lib64/lib64/libglfw.so.3 (0x00007f23b7d82000)
    libXrandr.so.2 => /usr/lib64/lib64/libXrandr.so.2 (0x00007f23b7b78000)
    libXxf86vm.so.1 => /usr/lib64/lib64/libXxf86vm.so.1 (0x00007f23b7972000)
    libXcursor.so.1 => /usr/lib64/lib64/libXcursor.so.1 (0x00007f23b7767000)
    librt.so.1 => /usr/lib64/lib64/librt.so.1 (0x00007f23b755f000)
[...]

PR #374, opened 2015-01-10
Fix win32 cmake generator without glew

hsdk:
- changed glew from a required to an optional dependency


manuelk:
In this particular case, i believe GLEW is actually required - can i ask why you think this is not the case ?

hsdk:
The required flag prevents generating the build files (see https://github.com/PixarAnimationStudios/OpenSubdiv/issues/358,  https://github.com/PixarAnimationStudios/OpenSubdiv/issues/373)  when OpenGL is found but GLEW is not available. With the change above, the build files are generated with and without GLEW. The rest of the cmake build script already treats GLEW as optional by checking for GLEW_FOUND before enabling the gl parts of the lib and the examples.

manuelk:
GLEW is only optional with OS X - this may not be the case any longer, but originally, i believe that we needed it for Win & Linux, which is why the REQUIRED flag is selectively added outside of OS X.

There is a separate flag to disable OpenGL entirely, and then GLEW is not a requirement any more. Would that solve your problem ?

manuelk:
Side note : please submit pull requests against the dev branch, not the master branch.

hsdk:
Sorry for submitting the pull request against the master branch. The intention was to build OpenSubdiv out of the box on systems that have OpenGL but not GLEW. On these system, the flag to disable the OpenGL parts is currenly mandatory to build the lib instead of an option. 

The GL parts of the library require both OpenGL and GLEW and this is not changed by the patch.

hsdk:
Sorry for submitting the pull request against the master branch. Shall i open a new request against the dev branch or just close this one?

manuelk:
Not necessary - i can cherry-pick the change manually. Let's finish hashing this out first as i think i am still missing something...

Your change essentially makes GLEW optional any time GL is found, which AFAIK causes problems on Win & Linux where GLEW is mandatory.

jcowles:
Closing -- Henry, please feel free to reopen if you're having this issue with the 3.0 dev branch!

PR #375, opened 2015-01-10
Minor gregory patch shader optimizations

hsdk:
- reduces instruction count by ~1/5
 - improves readability


manuelk:
thanks for the fix - we actually have some plans to shift these computations to a stencils basis, which should be much more efficient, so this section of the shaders is likely going to disappear in the not too distant future...

PR #376, opened 2015-01-11
Fix for GLTransformFeedback crash in computeStencil

hsdk:
- changed loop counter names in addWithWeight and readVertex that have overwritten the outer loop counter due to unrolling
- changed sizes sampler type to match GL_R8UI internal format

The crash occured on an NVidia 780, driver version 347.09, Windows. On Ati everything was fine.


manuelk:
Yikes - that's some impressive sleuthing skills there. I'll test this when i get back in the office and merge it then : thanks !

PR #379, opened 2015-01-13
Optimize GL/DX compute dispatches.

hsdk:
Minor optimization to match the number of needed workgroups to the batch size.

Example: 
```
int workgroupsize = 64; 
int batchSize = 64;

//old:
numDispatches = batchSize/workgroupSize +1;
// = 64/64+1 = 2;
// will start 2 workgroups - 64 threads that have to be setup but will imediately exit

//new:
int numDispatches = (batchSize+workgroupSize-1)/workgroupSize;
// = (64+63)/64 = int(1.98);
// will do what we want and start 1 workgroup
```


takahito-tejima:
cool fix, I like it!

PR #396, opened 2015-02-12
fix max valence issue for the case that only has one triangle face

shengfuosd:
When a mesh only has single triangle face, current code looks for the maximum valence in level 0, so max valence is 3. However, if the face is subdivided twice, the maximum valence is 4. This pull request is to fix this issue.

The directory layout for ICC library keeps changing in different releases. So I think it's much easier to use ICC_LIB_LOCATION, instead of ICC_LOCATION to find ICC related libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,8 +182,7 @@ if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANGCC OR CMAKE_COMPILER_IS_IC
                 NAMES
                     ${ICC_LIB}
                 PATHS
-                    ${ICC_LOCATION}/lib
-                    /opt/intel/lib/
+                    ${ICC_LIB_LOCATION}
                 PATH_SUFFIXES
                     ${ICC_LIB_ARCH}
             )
--- a/opensubdiv/far/patchTablesFactory.cpp
+++ b/opensubdiv/far/patchTablesFactory.cpp
@@ -696,6 +696,8 @@ PatchTablesFactory::createUniform( TopologyRefiner const & refiner, Options opti
         firstlevel = options.generateAllLevels ? 0 : maxlevel,
         nlevels = maxlevel-firstlevel+1;
 
+    if(maxvalence < 4) maxvalence = 4;
+    
     PatchDescriptor::Type ptype = PatchDescriptor::NON_PATCH;
     if (options.triangulateQuads) {
         ptype = PatchDescriptor::TRIANGLES;
@@ -843,7 +845,8 @@ PatchTablesFactory::createAdaptive( TopologyRefiner const & refiner, Options opt
     //  the inventory of patches determined above:
     //
     int maxValence = refiner.getLevel(0).getMaxValence();
-
+    if(maxValence < 4) maxValence = 4;
+    
     PatchTables * tables = new PatchTables(maxValence);
 
     // Populate the patch array descriptors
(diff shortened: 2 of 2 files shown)


manuelk:
Thanks for the fixes. I have a couple of minor issues though:

*  let's move the CMakeList changes into a separate issue. There are a couple of other details i would like to change right in that code, so i think i will file an issue and take a stab at fixing it if that's ok with you.
* can we fix the maxvalence lines with this instead ?
  ```
   int maxvalence = std::max(refiner.getLevel(0).getMaxValence(),4),
   ```
If this is agreeable, i would suggest closing this pull request and firing off a fresh one with just the change for the maxvalence.

Thanks again !

barfowl:
I would suggest alternatives for the maxValence() test.  The refined levels
are increasing the max valence accordingly, so testing the deeper levels
rather than level 0 may be more appropriate.

Alternatively, initializing the max valence internally with the minimum for
the applied scheme would spare everyone the need to externally do the
std::max with the regular valence (which is 6 for Loop rather than 4 for
Catmark).

- barry

manuelk:
Maybe add a public accessor to TopologyRefiner like this:
```
int GetMaxValence(int level=1);
```
If no level is specified, then it returns the hierarchical max valence. PatchTablesFactory should then be switched to use this function instead.

shengfuosd:
It seems make sense to add function getMaxValence in topologyRefiner, and implemented like the following:

int GetMaxValence() {
    return _levels[_maxLevel].getMaxValence();
}

barfowl:
I wouldn't presume that the _maxLevel always has the maximum valence.  If
TopologyRefiner wants to expose what it views as a collective maximum
valence, it can maintain its own member and internally adjust it as levels
are added.

- barry

manuelk:
I am also partial to having the ability of querying the max valence for a specific level. I would propose something like this:
```
int GetMaxValence(int level=-1) const;
...
 int
TopologyRefiner::GetMaxValence(int level=-1) const{
    int maxValence =0;
    if (level<0) {
        for (int i=0; i<GetMaxLevel(); ++i) {
            maxValence = std::max(_levels[i].getMaxValence());
        }
    } else if (level<GetMaxLevel()) {
        maxValence = _levels[level].getMaxValence();
    }
    return maxValence;
}
```

shengfuosd:
Looks good to me.

manuelk:
Barry is on his way to my office : i'll have his eyes on this shortly, and if there is consensus, we can move forward with the fix.

barfowl:
We only talked briefly about this yesterday, but a "level" argument is
passed to many methods of TopologyRefiner and I'm not aware of any cases
where the value of -1 is interpreted to apply the method to the object as a
whole.  For instance we have:

    int GetNumFacesTotal() const;
    ...
    int GetNumFaces(int level) const;

So if we want a max-valence query both for a specific level and for the
collection of levels, I suggest we follow the same pattern:

    int GetMaxValenceTotal() const;  //  we can probably drop the "Total"
here
    ...
    int GetMaxValence(int level) const;

Right now these collective queries don't have a result cached in a member
for immediate retrieval, but that's a simple improvement that can be made
if warranted.

- barry

barfowl:
I just came across this request -- thought it had been dealt with previously.  I can say with certainty that it has now.

In the spirit of my last comment, the TopologyRefiner had new members added to accumulate some of these "totals" or other properties that span multiple levels.  Each time a level is added, the properties are updated.  So TopologyRefiner now has a cumulative "max valence" member and a method to retrieve it:  GetMaxValence().

The PatchTables were updated to use this method and so should be getting the desired maximum they needed.

- barry

PR #399, opened 2015-03-30
osd_regression test ported over to new API and small bug fix in vtr

c64kernal:
vtr:

- Fixed a potential stack corruption issue with _maxEdgeFaces.

osd_regression:

- I had to disable the loop tests

- I had to disable the hedit tests

- Otherwise everything passes on Linux and Windows with all the existing tested backends (CPU, CPUGL, CL) -- we should add the rest.  (I couldn't test CL on Windows).

- The top-level CMakeLists.txt file still has osd_regression disabled because I haven't yet verified that it compiles on the Mac (it compiles on Windows and Linux).

- I think we should delete far_regression.  It's very outdated and the functionality is covered by vtr_regression.

### DIFF
--- a/examples/common/clInit.h
+++ b/examples/common/clInit.h
@@ -36,6 +36,7 @@
 #include "osd/opencl.h"
 
 #include <cstdio>
+#include <string>
 
 static bool HAS_CL_VERSION_1_1 () {
 #ifdef OPENSUBDIV_HAS_OPENCL
--- a/examples/common/clInit.h
+++ b/examples/common/clInit.h
@@ -59,6 +59,11 @@ static bool HAS_CL_VERSION_1_1 () {
 }
 static bool initCL(cl_context *clContext, cl_command_queue *clQueue)
 {
+    if (!clGetPlatformIDs) {
+        printf("Error clGetPlatformIDs call not bound.\n");
+        return false;
+    }
+
     cl_int ciErrNum;
 
     cl_platform_id cpPlatform = 0;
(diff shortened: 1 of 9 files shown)


barfowl:
Hey George, out of curiosity, why did you have disable the Loop tests?

And rather than deleting far_regression and relying on vtr_regression,
since vtr_regression uses Far classes, and Far is the public API (not Vtr),
why don't we repackage the current vtr_regression as far_regression?  I
don't see any direct use of the Vtr classes, just Far classes.  The "Vtr"
label is really only present to distinguish it from Hbr.

- barry

c64kernal:
Thanks Barry.  Yeah, I'm totally okay with rebranding vtr_regression to far_regression (note I didn't remove far_regression in these commits).

As for disabling the loop tests, it was just because they didn't work immediately (along with the hedit tests).  We have to add a task to re-enable them -- but I was hoping to give folks a head start with these tests by having osd_regression compilable in time for the next sprint.

jcowles:
Nice! 

All the osd_regression tests pass, however I get the following warning spew on OSX:

```
common/clInit.h:62:10: warning: address of function
      'clGetPlatformIDs' will always evaluate to 'true' [-Wbool-conversion]
    if (!clGetPlatformIDs) {
        ~^~~~~~~~~~~~~~~~
common/clInit.h:62:10: note: prefix with the address-of operator to
      silence this warning
    if (!clGetPlatformIDs) {
         ^

regression/osd_regression/main.cpp:299:31: warning: comparison of integers of different
      signs: 'size_type' (aka 'unsigned long') and 'int' [-Wsign-compare]
    assert(coarseverts.size() == refiner->GetNumVerticesTotal());

examples/common/clInit.h:41:13: warning: unused function
      'HAS_CL_VERSION_1_1' [-Wunused-function]
static bool HAS_CL_VERSION_1_1 () {
```

I could fix this after the merge, if you prefer.

c64kernal:
Thanks so much, Jeremy.  Yeah, if you wouldn't mind fixing it after the merge, that would be awesome.  Also, now that you've verified it compiles on OSX, please feel free to enable building osd_regression by default in regression/CMakeLists.txt.  Thanks again!

jcowles:
Ok, fixed in 28e39bf

PR #400, opened 2015-03-31
Alternative fix to clInit.h for cross-platform.

c64kernal:
An alternative fix to clInit.h that should allow the Windows build to conitnue working.

### DIFF
--- a/examples/common/clInit.h
+++ b/examples/common/clInit.h
@@ -57,16 +57,20 @@ static inline bool HAS_CL_VERSION_1_1 () {
     return false;
 #endif
 
+#ifndef _WIN32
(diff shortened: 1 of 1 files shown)


jcowles:
Doh. Ok, we need to find the correct fix for this. It seems in some cases (Clew related?) we are getting function pointers and sometimes not. I'm curious why this function isn't bound on your machine, maybe if we get to the bottom of that first, then we can roll out the correct fix.

jcowles:
This is definitely a result of CLEW, I have a fix that makes this conditional based on its presence.

c64kernal:
Much nicer fix, thanks Jeremy!

PR #424, opened 2015-04-18
Refactor table factory APIs

takahito-tejima:
This change moves all gregory patch generation from Far::PatchTablesFactory so that we can construct patch tables without stencil tables as well as client can chose any end patch strategies (we have 2 options for now: legacy 2.x style gregory patch or gregory basis patch).
Also Far::GregoryBasisFactory provides index mapping from patch index to vtr face index, which can be used for single gregory patch evaluation on top of refined points, without involving heavier stencil tables.


takahito-tejima:
I'll make another request.

PR #430, opened 2015-04-19
Fixing FindGLFW cmake module

jcowles:
Previously, FindGLFW would search for both GL/glfw.h and GLFW/glfw3.h, which
does not work when a system has both GLFW 2.0 and 3.0 installed.

The fix still probes for both files, but does so in two passes, preferring
glfw3.h.


takahito-tejima:
great, thanks.
I think we can drop glfw 2.0 support at some point.

PR #433, opened 2015-04-21
Separating out Ptex API from TopologyRefiner.

c64kernal:
Remove the ptex-specific code from the Far::TopologyRefiner and instead provide it in a separate class Far::PtexIndices.  Clients who need to use the Ptex API need to first build a Far::PtexIndices object by providing it with a refiner.

This has the advantage of keeping the API on the TopologyRefiner a little cleaner.  The ptex methods were const but were mutating state with const_casts.  The new mechanism still achieves the same lazy initialization behavior by forcing clients to instantiate them exactly when needed.

A disadvantage of this approach is that the PatchTablesFactory creates its own PtexIndices and throws it out after the patch tables are created.  This is great if you're never going to need the ptex indices again, but not so great if you will need them again.


jcowles:
Few minor comments:
 * Would be nice to add a doc string for the class (right now it's just "\brief")
 * We could remove the "Ptex" name around the body functions, since this class is exclusively about ptex
 * Instead of an assert(scheme=catmark), should we instead check the condition, emit and error (see far/error.h) and bail early? The assert will go away in release code and then clients may hit undefined behavior with no warning.
 * The copyright on the new file should be Pixar (the standing rule is whoever creates the file gets the copyright)
 * Making more friends every day! (we really need to fix this throughout the code base)

c64kernal:
@jcowles thanks for the review!

- doc string: Yup, sorry I missed that, I meant to come back to that if we were going to accept this.  Fixed.

- remove "Ptex" name from body functions.  I'd like to reconsider.  The word "Index" gets thrown around easily.  For example GetPtexIndex() would become GetIndex() and it takes an Index.  It can get pretty confusing.  I like the explicitness.  If we formalize the parameterization API that might be a better time to change the names?  What do you think?

- assert.  Unfortunately that's the pattern everywhere.  Since this function is potentially inner-loop stuff, I wonder if the debug-only check is on purpose.  One possibility is to add such a check to the constructor so that at least you can get a nice error message once there and avoid surprises later on?

- Copyright notice.  Fixed.

- Friends: yeah, agreed, we really do have to address that throughout.

barfowl:
Good point about removing "Ptex" from "PtexIndex".  I can see using "Face"
in place of index -- the index is to a Ptex face after all, and we "Get"
other indices for things like face-vertices without referring to "Index" at
all in the method name.  But if we were to refer to a PtexFace that still
has Ptex in the name -- the distinction may be worth it.

I favor putting the error checking in the constructor in place of the
asserts everywhere.  Does Ptex apply to faces from Bilinear subdivision?
If so, you can test if the scheme generates quads as follows:

    if (Sdc::SchemeTypeTraits::GetRegularFaceSize(refiner.GetSchemeType()
== 4) ...

I try to avoid testing against specific scheme types when possible.

- barry

c64kernal:
I'm okay with renaming GetPtexIndex() to GetPtexFace().  Let me know if I should go ahead and do that.

I'll go ahead and add the error check in the constructor and I'll keep the assert at the top of the method, but replace it with the traits check.  The reason I'd like to keep it in the method is because of the choice to pass in the refiner and not keep it referenced, also it shouldn't cost anything in release builds.

Thanks!

barfowl:
Sounds good -- you can add the traits test to the "sanity check" I
mentioned to validate the Refiner that's passed in.

I defer to the rest of you regarding GetPtexFace().  Being a relative PTex
newbie there may be a compelling case to keep Index -- the class is called
PtexIndices after all.

- barry

jcowles:
We should peek at the ptexture api, will do once I get home.

--
Jeremy

Sent from my phone

c64kernal:
Error checking in place now.

The ptexture API appears to refer to it as faceid (good call on checking there).

jcowles:
Ah yeah, faceid and describes it as a "face index" in the docs.

barfowl:
Ah, now for a really important decision:  GetFaceid() or GetFaceId()?

- barry

c64kernal:
Here's a proposal:

GetNumPtexFaces() --> GetNumFaces()
GetPtexIndex() --> GetFaceId()
GetPtexAdjacency() --> GetAdjacency()

jcowles:
I like it! The entire class is already in the context of "PtexIndices", no need to repeat Ptex everywhere.

c64kernal:
Sold!

davidgyu:
Cool. I like this too!

PR #435, opened 2015-04-21
Limit position screen-space tessellation level metric

davidgyu:
Added a new screen-space adaptive tessellation metric which
computes edge lengths using limit surface points. Made this
the default screen-space metric so that we avoid cracks when
using Gregory Basis or Regular B-spline end caps.

The alternative method which computes edge lengths using the
distance between B-spline control points is still available.

Added a diagram and comments to explain how the control
points and limit points are organized.

### DIFF
--- a/opensubdiv/osd/glslPatchBSpline.glsl
+++ b/opensubdiv/osd/glslPatchBSpline.glsl
@@ -85,7 +85,7 @@ out block {
 #define ID gl_InvocationID
 
 void
-reflectBoundaryEdges(inout vec3 cpt[24], int patchParam)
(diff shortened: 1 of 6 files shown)


jcowles:
Looks great! @davidgyu and I sat down and reviewed this, but it would be good to get @takahito-tejima 's eyes on it for a post review also.

takahito-tejima:
It looks great.
EvalBezier for tess metric could be simplified since we only need it at 0.5 parameter of 4 edges (no tensor product), but maybe shader compiler optimizes all out and gets the same result?

awesome change, thanks David!

PR #436, opened 2015-04-21
Fixes far_tutorial_6 crash

takahito-tejima:
following recent API changes, specifiy endcap factory.

### DIFF
--- a/tutorials/far/tutorial_6/far_tutorial_6.cpp
+++ b/tutorials/far/tutorial_6/far_tutorial_6.cpp
@@ -42,6 +42,7 @@
 
 #include <opensubdiv/far/topologyRefinerFactory.h>
 #include <opensubdiv/far/patchTablesFactory.h>
+#include <opensubdiv/far/endCapGregoryBasisPatchFactory.h>
 #include <opensubdiv/far/patchMap.h>
 
 #include <cassert>
@@ -154,8 +155,10 @@ int main(int, char **) {
 
     // Generate a set of Far::PatchTables that we will use to evaluate the
     // surface limit
+    Far::EndCapGregoryBasisPatchFactory endcapFactory(*refiner);
     Far::PatchTables const * patchTables =
-        Far::PatchTablesFactory::Create(*refiner);
+        Far::PatchTablesFactoryT<Far::EndCapGregoryBasisPatchFactory>::Create(
+            *refiner, Far::PatchTablesFactory::Options(), &endcapFactory);
 
     // Create a Far::PatchMap to help locating patches in the table
     Far::PatchMap patchmap(*patchTables);


c64kernal:
Thanks for the fix, Takahito.   Is there anything we can do to make this be a compile-time error rather than a run-time crash or at least some kind of error to help clients not trip over this?

-George & Jeremy

takahito-tejima:
Good point, since adaptive refinement usually requires some kind of end patch, options would be
1. make legacy gregory patch as default end patch (even for the users use uniform only)
  or 
2. strictly distinguish uniform and adaptive refinement (by patch factory type). Then client can't compile adaptive table factory without specifying end patch.

We'd like to recommend to use gregory basis or regular endcapping in 3.0 for various reasons, however, both requires additional stencils which may affect client design so I don't think we can make them default at least at this point. Also hopefully we want to retire legacy evaluation at some point, so option 1 doesn't sound good to me.

option2 gives more strict compile time check, but we need to change PatchTablesFactory::Create function to CreateUniform() and CreateAdaptive(). I'm ok with that but the comment in PatchTablesFactory says:

 881 //  We should be able to use a single Create() method for both the adaptive and uniform
 882 //  cases.  In the past, more additional arguments were passed to the uniform version,
 883 //  but that may no longer be necessary (see notes in the uniform version below)...

so there might be a reason we want to use same interface on PatchTableFactory to create both uniform and adaptive patches, whereas uniform case we don't need endcap factory.

jcowles:
Could it be somehow embodied in the Options?

--
Jeremy

barfowl:
I may have been the one who wrote that comment about the single Create()
method.  I've always felt we shouldn't try to specialize the choices too
much based on the way the mesh was refined.  In future there may be more
flexible kinds of sparse refinement -- not necessarily feature adaptive --
and I see most of the choices we have now applying to that case too.  Right
now a uniformly refined mesh dictates bilinear patches, but I've argued
before that there's no reason not to provide smooth patches (yes, that
usage can be abused, but so can uniform refinement in general).

So I wouldn't be in favor of (2).

If a uniformly refined mesh is to have bilinear patches, unless otherwise
specified, I would expect linear endcaps.  If an option was to specify
bicubic patches, I would expect some kind of smooth endcap.  But that
doesn't really address the issue here.

Why is the EndCapFactory being passed to the PatchTablesFactory rather than
being created internally by it as needed?  Is there any use for it outside
the patch table construction?

- barry

takahito-tejima:
Currently EndCapGregoryBasisPatchFactory is used after patchtable creation to get vtrlevel from patchtalbe, then we can evaluate gregory basis patch without stencil tables.
If we store vtrLevel into somewhere else indexed by patchIndex (like FarPatchParam), we don't need to hold EndCapFactory. I added an example usage for GregoryBasis evaluation without stenciltables case in vtrViewer, but my usage is always with stencils so I'm not sure if it makes sense or not...

barfowl:
So all of the regular patches have gathered their vertices from the Level
during patch construction and stashed them within the PatchTables, but the
Gregory basis patches do not -- and so subsequently need to gather them
later?  What else do you need from the Level other than its points?

If that's true, is there a reason we can't gather everything that we need
for the Gregory basis patches like we do for the others and store it in the
PatchTables in a new member?  Each Gregory basis patch would have a
different number of CV's and need size/offset to its points.  Are we trying
to avoid storing all of this?

(I should probably look more at what's going on before saying much more --
I feel like I'm missing something...)

- barry

takahito-tejima:
I see, let me summarize my thoughts. At first I apologize for the lengthy response :)
 
Currently we have three options (factories) for the endcap patch generation. PatchTablesFactory delegates endcap patch topology population to one of those factories which specified as template parameter. We'll add more factories later (just quad cap, DFAS, etc). The basic idea here is PatchTablesFactory doesn't need to know that the endcaps are gregory or regular spline or quad.

1. EndCapLegacyGregoryPatchFactory
This factory follows OpenSubdiv 2.x pattern. It generates 4 vertices zero-ring into the patch table and also populate vertex valence table (aligned at max-valence) and quad offsets table. All information is stored into PatchTable, so the subsequent process needs nothing else. Osd eval and draw API take them and create an inflight Gregory patch (20 verts) every time. This is inefficient and complicated, so we'd like to retire this scheme soon in 3.0.

2. EndCapGregoryBasisPatchFactory
If this class is specified as an endcap factory, this factory gives 20 control points (boundaries could be shared across adjacent gregory patches) back to PatchTablesFactory, as the factory accumulates the stencils for those new vertices by gathering vertices. Once the PatchTablesFactory is done, this factory gives a vertexStencilTable for the additional gregory patch vertices, either factorized or not factorized to the regular refined vertices stencils.

If we don't use regular vertex stencils, this factory gives relatively small stencilTables, and we can compute gregory vertices from the max level of refined vertices. So no vertex gathering is needed in these subsequent processes (However since PatchTablesFactory is statelss we need a container or something for the returning stencils anyway).

If we don't even want to use such additional stencils, inflight gregory patch can be obtained by gathering vertices, which is the same way as above stencil accumulation. This is where I haven't understood well the actual use cases, but I assume we have a topologyRefiner since we interpolate max-level regular vertices without stencilTables, so I think it woud be reasonble to ask topologyRefiner to gather vertices for that inflight Gregory patch generation, then we'll need vtr face index there. But maybe I'm wrong.

>If that's true, is there a reason we can't gather everything that we need for the Gregory basis patches like we do for the others and store it in the PatchTables in a new member?

No I think it makes sense, but that is almost what EndCapLegacyGregoryPatchFactory does (except memory inefficiency). We can improve data structure of VertexValenceTable and QuadOffsetsTable, however, when we'll start to implement DFAS we'll likely need multi-level topology traversal around extraordinary vertices, which sounds topologyRefiner's (or VtrLevel's) role very much, I think... and even so, it's not a problem as long as the EndCapFactory uses topologyRefiner within the context of PatchTablesFactory and it generates stencils. I feel like using small stencils is simpler and easier than encoding topolgy into patch table.

3. EndCapRegularPatchFactory
This is still experimental and off-topic, but extremely useful for unified shader GL drawing as all patches become regular bspline. This factory works similar as EndCapGregoryBasisPatchFactory, but even more approximates gregory patch to bezier patch, and applies bezier to bspline conversion to get 16 control points, populates them into the patch tables.

takahito-tejima:
David and I discussed about this and we have an idea that how to generalize end capping and to get more better and safer API. Give me some time to try it. thanks!

barfowl:
Okay, thanks.  From a users perspective I have always envisioned that the
PatchTablesFactory takes a Refiner and some options, and all of the
complexity required for construction takes place "under the hood".  That's
why the exposure of additional factory classes that users have to
understand and manage concerns me.

I've been sitting on some other ideas about PatchTables, patches in general
and their construction, and maybe its time I put those out there for
consideration.  Nothing revolutionary or contrary to what's going on, but
it may shift the class design away from some of the lower level factories
we're considering, while potentially giving users lower level access to
patches without full construction of PatchTables.  Now that I think of it,
if we do expose lower level Factory classes as part of the public API for
constructing PatchTables, that will complicate things if we decide later we
no longer need them, so keeping their use out of the public API would be
preferable.

I was planning on waiting until after 3.0 when we start to revisit
face-varying patches and start to consider patches for Loop, but if guys
have the time to consider them, I might as well put them out there sooner
than later.

So look for a lengthy new email thread related to PatchTables coming soon
-- I've got a production issue I need to look into and will try to get that
out as soon as I can.

- barry

jcowles:
I'm going to merge this to fix the immediate breakage, then we can circle back revisit the larger design.

PR #440, opened 2015-04-24
retain vertex 0 for lefthanded meshes

asluk:
Keep vertex 0 as the start of the winding order for lefthanded meshes; scale down left/righthanded mesh examples to avoid precision issues, and add the lefthanded one to the far regression suite.


takahito-tejima:
great, thanks!

PR #441, opened 2015-04-24
Fixed orientation of boundary and corner patches

davidgyu:
Now a boundary and corner patch remains
aligned with its underlying parametric
orientation. This simplifies both the
gathering of patch vertices and downstream
evaluation of patches.

Added a method to Vtr::Level which gathers
the 16 patch points for a B-spline patch
even if the patch has boundary or corner
edges.  The undefined patch vertex index
values along boundary and corner edges are
assigned Vtr::INDEX_INVALID.

In order to simplifiy the process of drawing
B-spline patches with boundary or corner
edges, the Far::PatchTablesFactory will
replace any invalid vertex indices with
a known good value, i.e. the index of the
first patch face vertex.

Single-crease patches are still a slightly
special case, which will be resolved later.

### DIFF
--- a/opensubdiv/far/patchTablesFactory.cpp
+++ b/opensubdiv/far/patchTablesFactory.cpp
@@ -182,20 +182,32 @@ offsetAndPermuteIndices(Far::Index const indices[], int count,
                         Far::Index offset, int const permutation[],
                         Far::Index result[]) {
 
+    // The patch vertices for boundary and corner patches
+    // are assigned index values even though indices will
+    // be undefined along boundary and corner edges.
+    // When the resulting patch tables are going to be used
+    // as indices for drawing, it is convenient for invalid
+    // indices to be replaced with known good values, such
+    // as the first un-permuted index, which is the index
+    // of the first vertex of the patch face.
+    Far::Index knownGoodIndex = indices[0];
+
     if (permutation) {
         for (int i = 0; i < count; ++i) {
-            if (permutation[i] < 0) { // XXXdyu-patch-drawing
-                result[i] = offset + indices[0]; // XXXdyu-patch-drawing
-            } else {
-                result[i] = offset + indices[permutation[i]];
+            Far::Index index = indices[permutation[i]];
+            if (index == Vtr::INDEX_INVALID) {
+                index = knownGoodIndex;
             }
-        }
-    } else if (offset) {
-        for (int i = 0; i < count; ++i) {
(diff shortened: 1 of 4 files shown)


takahito-tejima:
David and I looked through this change together. This is a really nice cleanup.
I expect fvarchannel patch generation will follow the same simplification too. exciting.

takahito-tejima:
I just noticed non-manifold case (catmark_fan) is broken. Both boundary and corner case, vFaces.size() may not be 1 or 2. Probably using boundaryMask and rotation to find undefined verts would be better?

davidgyu:
Oh, I'll take a look. Thanks!

barfowl:
I haven't looked at this in detail -- though I did see the new gather
method in Vtr::Level that fills in the full 16 points -- but may have some
information to help deal with non-manifold patches...

The original Vtr::Level patch gathering methods -- specifically the ones
for boundary and corner patches -- do not make any assumptions about the
topology beyond exactly what they are told to gather.  So these can
(intentionally) be used to gather partial B-spline patches for faces that
have non-manifold vertices or edges.  For a patch in the catmark_fan case,
we know that for those faces along the non-manifold edge that we need to
gather boundary patches, and we never deal with that non-manifold edge or
its non-manifold vertex as part of the gathering.

The same is true for face-varying gathering -- the "boundaries" or
"corners" (face-varying discontinuities in this case, not topological ones)
are already known and there is nothing conditional on the topology as part
of the actual gathering within these methods.

I'm wondering if we should modify those original boundary and corner
methods to pad the missing patch points to 16 rather than having a single
method that tries to do all.  Otherwise its specification will need to
identify rotation, boundary edges and/or corner vertices all in one.

Is any part of this work intending to lay the groundwork for supporting
regular patches with multiple boundaries or corners?  Right now the
gathering assumes a single boundary or corner and all other vertices are
regular interior.

- barry

davidgyu:
Downstream drawing and evaluation client code still requires that features have been isolated, so no real change in assumptions.

My extension to the gather interface was overly optimistic, and the client code (patchTablesFactory) does know the topological restrictions that have already been discovered and recorded in the tags.

I think it would be fine to either use the pre-existing gather interfaces as they were, and have the patchTablesFactory expand and rotate (unrotate) the boundary and corner patch points, or to extend the boundary and corner gather methods to support padding and rotating (unrotating) the patch points.

-David

PR #442, opened 2015-04-24
Fix ptex indices computations for loop.

c64kernal:
The only method not yet implemented for loop in PtexIndices is GetAdjacency(), fix code accordingly.


davidgyu:
Looks good!

PR #445, opened 2015-04-27
Fix an out-of-bounds read.

c64kernal:
Fixed a bad read from a std::vector.  It's normally harmless, but at least makes
debug runtimes cry.


takahito-tejima:
Thanks for fixing!

PR #446, opened 2015-04-27
Temporary fix for GL version detection in glViewer (issue #358)

c64kernal:
The way we ask for a GL version leaves room for improvement.  In this change I only changed it so that we test for the minor version at runtime rather than what we can grep out of the glew headers at compile time.  The full desired solution described in issue #358 provides a more complete solution.

This change makes glViewer work on machines that only support OpenGL 4.2 but
does not provide a nicer fix than that.


c64kernal:
This doesn't work as it puts the cart before the horse.

PR #448, opened 2015-04-27
Fixed gathering of patches near non-manifold topo

davidgyu:
The previous change to the gathering of patch points went
a bit too far. Near non-manifold features it is important
to be careful when traversing the faces in a level to avoid
assumptions that are valid only for manifold topology.

Also, removed Vtr::Level::gatherQuadRegularPatchPoints().
This method was added in my previous change, but it is
unsafe to use in the presence of non-manifold topology.


takahito-tejima:
I merged this request and confirmed locally that it no longer crashes on catmark_fan.
Thanks for the fix!

PR #449, opened 2015-04-28
cleanup examples to use common/clDeviceContext.h, cudaInit.h

takahito-tejima:
move CL/CUDA specific initialization stuffs into
common/clDeviceContext.{h,cpp} and common/cudaInit.h
so that they can be referred from both examples and regression tests.

update examples to use same CLDeviceContext struct.
fix glShareTopology drawing bug.
(still something strange with XFB kernels)

Note that topdir/common is meant to be used just for examples and regression tests,
so don't include a file in common from opensubdiv lib sources.
(do we like to rename /common to /utils or something?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -688,6 +688,8 @@ endif()
 
 add_subdirectory(opensubdiv)
 
+add_subdirectory(common)
+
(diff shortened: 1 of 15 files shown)


takahito-tejima:
Closing to reconsider directories.

PR #450, opened 2015-04-28
cleanup CL/CUDA example harnesses.

takahito-tejima:
refactor CL/CUDA specific initialization stuffs into
examples/common/clDeviceContext and cudaDeviceContext, and
update examples to use those structs.

also
- remove CL/CUDA tests from osd_regression. The tests for those
  kernels will be covered by glImaging.
- update cuda initialization to use the GL-interoperable device
  if available.
- remove CL specialization from glShareTopology, following the same
  pattern as we took in the previous OsdGLMesh refactoring.
    (still something strange with XFB kernels though)
- fix file permissions.


takahito-tejima:
Closing for further fix.

PR #461, opened 2015-05-01
Satisfy compiler warnings about unused parameters.

c64kernal:
In vtr/stackBuffer.h, removed  unnecessary symbols for parameters to silence
compiler warnings about them not being used.

### DIFF
--- a/opensubdiv/vtr/stackBuffer.h
+++ b/opensubdiv/vtr/stackBuffer.h
@@ -73,7 +73,7 @@ public:
 private:
     //  Non-copyable:
     StackBuffer(const StackBuffer<TYPE,SIZE> &) { }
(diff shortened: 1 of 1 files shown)


barfowl:
My bad, thanks.  Looks like operator= should also be returning *this
instead of just { }.

- barry

c64kernal:
Good call.  Coming right up!

barfowl:
Thanks.  Its surprising that whatever compiler was warning about the unused
parameters didn't also complain about the missing return value...

- barry

c64kernal:
I was thinking the same thing...

takahito-tejima:
Thanks, now the travis build log looks very clean :)
https://travis-ci.org/PixarAnimationStudios/OpenSubdiv/builds/60787450

PR #463, opened 2015-05-06
Updated handling of patchParam and patchCoord

davidgyu:
Each patch has a corresponding patchParam. This is a set of three values
specifying additional information about the patch:

   faceId    -- topological face identifier (e.g. Ptex FaceId)
   bitfield  -- refinement-level, non-quad, boundary, transition, uv-offset
   sharpness -- crease sharpness for a single-crease patch

These are stored in OsdPatchParamBuffer indexed by the value returned
from OsdGetPatchIndex() which is a function of the current PrimitiveID
along with an optional client provided offset.

Accessors are provided to extract values from a patchParam. These are
all named OsdGetPatch*().

While drawing patches, the patchParam is condensed into a patchCoord which
has four values (u, v, faceLevel, faceId). These patchCoords are treated
as int values during per-prim processing but are converted to float values
during per-vertex processing where the values are interpolated.

Also, cleaned up more of the shader namespace by giving an Osd prefix
to public functions, and consolidated boundary and transition handling
code into the PatchCommon shader files. The functions determining
tessellation levels are now all named OsdGetTessLevel*().


takahito-tejima:
We reviewed this change together. very nice cleanup.

PR #464, opened 2015-05-07
Remove osd/vertex.h, osd/*ptex* classes

takahito-tejima:
OsdVertex was used in 2.x FarMesh as a placeholder for Hbr template parameter.
We don't need such a class any more.


davidgyu:
This looks great! Thanks!

PR #465, opened 2015-05-07
osd refactor: split the vertex buffer argument into src and dst

takahito-tejima:
Changing all device kernels to take two buffer identifiers for
source and destination separately.
This change is an intermediate step toward upcoming context/controller
refactoring.

Previously we have a limitation that the source and destination
vertex buffer have to be a single buffer, since the subdivision
tables are iteratively applied level by level.
With stencil tables, we don't have such a limitation any more,
so we may want to apply stencils from a separate source buffer to
another.

To specify the stencil output location within the destination buffer,
we use VertexBufferDescriptor.offset. This allows us not only
configuring arbitrary batching scheme, but also relaxing the
limitation that source and destination buffers had to have a same
interleaved layout. For examples, with this change we could layout
derivatives only exist in the destination buffer, which doesn't need
to be allocated in the source buffer.

### DIFF
--- a/opensubdiv/osd/clComputeController.cpp
+++ b/opensubdiv/osd/clComputeController.cpp
@@ -63,16 +63,22 @@ class CLComputeController::KernelBundle :
 
 public:
 
-    bool Compile(cl_context clContext, VertexBufferDescriptor const & desc) {
+    bool Compile(cl_context clContext,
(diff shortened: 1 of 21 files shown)


davidgyu:
This looks great! Laying a nice foundation for subsequent changes! -David

PR #466, opened 2015-05-09
Refurbish osd layer API.

takahito-tejima:
In OpenSubdiv 2.x, we encapsulated subdivision tables into
compute context in osd layer since those tables are order-dependent
and have to be applied in a certain manner. In 3.0, we adopted stencil
table based refinement. It's more simple and such an encapsulation is
no longer needed. Also 2.0 API has several ownership issues of GPU
kernel caching, and forces unnecessary instantiation of controllers
even though the cpu kernels typically don't need instances unlike GPU ones.

This change completely revisit osd client facing APIs. All contexts and
controllers are replaced with device-specific StencilTable and Evaluator.
While we can still use consistent API across various device backends,
unnecessary complexities have been removed. For example, cpu evaluator
is just a set of static functions and there's no need to replicate
FarStencilTables to ComputeContext.

Also the new API delegates the ownership of compiled GPU kernels
to clients, for the better management of resources especially in a multiple
GPU environment.

In addition to integrating ComputeController and EvalStencilController into
a single function Evaluator::EvalStencils(), EvalPatches() is also added
into Evaluator as a replacement of EvalLimit. This is still work in progress,
and we'll make a followup change for the complete implementation.

- some naming convention changes:
GLSLTransformFeedback to GLXFBEvaluator
GLSLCompute to GLComputeEvaluator

- move LimitLocation struct into examples/glEvalLimit.
We're still discussing patch evaluation interface. Basically we'd like
to tease all ptex-specific parametrization out of far/osd layer.

TODO:
- implments EvalPatches() in the right way
- derivative evaluation API is still interim.
- VertexBufferDescriptor needs a better API to advance its location
- synchronization mechanism is not ideal (too global).
- OsdMesh class is hacky. need to fix it.


davidgyu:
Takahito and I have reviewed this together. There are quite a lot of changes here, but this is a good step forward. -David

PR #478, opened 2015-05-15
Copy past error for colorMap variable

pkitslaar:
Small copy-past error which redefines colorMap instead of declaring normalMap variable.

### DIFF
--- a/examples/common/gl_framebuffer.cpp
+++ b/examples/common/gl_framebuffer.cpp
@@ -188,7 +188,7 @@ GLFrameBuffer::compileProgram(char const * src, char const * defines) {
     if (colorMap != -1)
         glUniform1i(colorMap, 0);  // GL_TEXTURE0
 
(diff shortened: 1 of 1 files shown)


jcowles:
Thanks! Do we have a signed CLA for you already? If so, what name/company should I look for?

Also, would you mind making this change against the dev branch and re-requesting?

If that's too much of a pain, it's small enough that we can just manually incorporate this fix for you -- thanks for pointing it out!

pkitslaar:
I don't have a signed CLA, where/how should I do that?

I will make the fix against dev and re-request.

jcowles:
Take a look here: http://graphics.pixar.com/opensubdiv/docs/intro.html#contributing

--
Jeremy

jcowles:
Looks like this is already fixed in the dev branch.
Closing.

pkitslaar:
Ok. 

Btw I am still waiting working on the CLA with my employer since I got some more patches in the future.

PR #480, opened 2015-05-15
update glShareTopology example.

takahito-tejima:
This example is rewritten as a more comprehensive example
of Far and Osd APIs to generate batched index buffer and
vertex buffer, as well as sharing same topology and stencil
table among multiple objects.

Also this change includes an experimental code path of using
glMultiDrawElementsIndirect. It's currently incomplete due to
the missing interface of osd tessellation shader.


davidgyu:
Takahito and I reviewed this together on Friday. Looks great and it's nice to see how well the API and client code are cleaning up!

PR #462, opened 2015-05-03
Dev

ielillo:
I've been working on making the glViewer work on older machines. From the initial executions the program failed because the wrong opengl version hinting was being set in the function setGLCoreProfile. In order to fix that I decided to try different opengl versions and fail if none of the works. The version are 4.4, 4.2 and 3.2.

Another issue was the shader linking errors. The cause was that the wrong opengl version  was being set in the shader file. As a solution, I created a function to query for the current opengl version and use it to set the right shader version before compiling it.

Finally some parts of the code used the GLEW headers definitions to check if certain opengl capabilities were present, the issue with that is sometimes a driver says that certain capability is present but it does not says that it is really implemented. So if GLEW was present I used it to dynamically query the opengl capabilities falling back to headers definitions otherwise. 

Regarding the platforms  this code has been developed primarily on Windows and Linux so it should work on both. However it has not been compiled nor tested on the Mac platform, so I cannot guarantee that it will compile there or work.
I'm open to any suggestion or criticism.

PR #484, opened 2015-05-19
remote SupportsAdaptiveTessellation from GLDrawContext (and example cleanups)

takahito-tejima:
As a preparation for retiring DrawContext, move SupportsAdaptiveTessellation
method to examples/common/glUtils, which is renamed and namespaced
from gl_common.{cpp,h} to be consistent to other files.
Same renamings applied to other example files.

SupportsAdaptiveTessellation is fixed to see runtime availability, as suggested in
#462 .


takahito-tejima:
Closing to resolve a conflict.

PR #488, opened 2015-05-19
Prefix header guards and add pragma once.

jcowles:


barfowl:
FWIW, ICC warns me (or rather "remarks"):

    remark #1782: #pragma once is obsolete. Use #ifndef guard instead.

Do we want to introduce an "obsolete" mechanism at this point?  It may
cause more noise for people than it helps anyone else.

- barry

jcowles:
Yeah, we debated this, particularly because "once" was actually struck down for standardization -- the hope was to safely speed up compile time by providing both pragma and the ifndef guard and I hoped that the worse case would be that they would be ignored.

I'm not sure I buy the fact that they are "obsolete", but I can back that bit of my change out.

--
Jeremy

barfowl:
Do you think or see that any other compilers are generating noise over
them?  It doesn't bother me enough to care that much about it -- its easily
suppressed -- so I don't object to it being left in if it has some use.

- barry

c64kernal:
I feel like we should back out the #pragma once since it's not standard and the OpenSubdiv codebase is not large enough to really benefit from an increase in compile time.  Just my gut feeling though.

barfowl:
The "not standard" point swings me towards removing it if we are
trying/claiming to be standard.

- barry

jcowles:
OpenSubdiv compile times can get pretty lengthy, but yeah, I still think we should back it out for the sake of being standard.

--
Jeremy

jcowles:
And I just realized, compile times may be long due to templates, but pragma once is perhaps unlikely to save time in the way that would be most effective.

I also think we should have first quantified the actual performance improvement, rather than rolling out an optimization and crossing our fingers!

--
Jeremy

PR #502, opened 2015-05-20
Refactor public patch evaluation methods exposed by Far::PatchTables

barfowl:
These changes eliminate Far::PatchTables Evaluate...<T,U>() methods -- replacing them with the un-templated EvaluateBasis() so that clients can get the weights they need directly and combine their control point data however they want.  This eliminate the remaining need for the Interpolate...<T,U>() methods that also needed to be publicly included via far/interpolate.h.

The remaining low-level basis evaluation code has been moved from far/interpolate.* to far/patchBasis.*.  This is now private/internal and we should be free to redefine and reorganize what we need (including second derivative access) in future.

This change warrants a little more testing than I could give it -- unless the Travis CI confirmation is adequate, it may warrant someone pulling these changes down and confirming them before the pull.


barfowl:
Looks liked I missed some usage of the methods I deprecated -- one in the FarViewer and another in far/tutorial_6.  Will see what's needed there and update...

barfowl:
Updated tutorials/far/tutorial_6 and disabled some face-varying usage in examples/farViewer.

The existing usage of EvaluateFaceVarying() does raise the issue of whether/how to generalize EvaluateBasis() for Varying, FaceVarying, etc. that was mentioned offline earlier today.

PR #506, opened 2015-05-21
Remove need for friending a class in PatchTables.

c64kernal:
PatchTables no longer needs to friend EndCapLegacyGregoryPatchFactory.
Instead we now make the patch table factory pass in the data that needs
to be updated directly to EndCapLegacyGregoryPatchFactory.


c64kernal:
Did not mean to do that.

PR #509, opened 2015-05-21
Raised maximum valence supported by Vtr and included new test shapes:

barfowl:
With the StencilTable size recently raised from char to int, we should now be set to raise the valence limit from the max of an 8-bit to 16-bit integer.  The LocalIndex typedef has been changed and new tests shapes in excess of the old 8-bit limit provided.

The new test shapes were added to regression/far_regression but were disabled.  Failures arose but due to the fact that Far's application of weights is numerically more accurate that Hbr's.  We'll be dealing with it passing the far_regression test at a later date.


jcowles:
Rather than hard coding 16 bits, should we use sizeof(short)?

barfowl:
We shouldn't be hard-coding anything to 16 bits -- unless its the valence
limit constant you're referring to.

It's the max that we need, and in the absence of being able to use
numeric_limits (static initialization in C++ is a pain) we would still need
(1 << 8*sizeof(short)) - 1.  We should really use LocalIndex here instead
of short if we were to go that route, and that's what I almost did but it
seemed unduly complicated.

I was tempted to go old-school and use the equivalent of MAXINT for short,
but I'm not sure that is standard, and it requires bringing in another
header file into one that is widely included and for only one unfortunate
usage, and its not parameterized by LocalIndex either.

I had a comment regarding numeric-limits next to the declaration -- we
should be able to use it thanks to constexpr in C++11 and replace whatever
is there now -- but at a certain point I decided I was just overthinking
this.

Anything other than std::numeric_limits<LocalIndex>::max() has its
drawbacks...

- barry

jcowles:
Yeah, that was what I was referring to. 

My worry is that someone would update one and not the other, but it's probably not likely to be updated often.

barfowl:
Exactly -- unless you can parameterize the second in terms of the first
you're forced to update both.  Hopefully a code review prevents anyone
updating one and not the other (in the unlikely event its needed).

- barry

PR #510, opened 2015-05-21
glStencilViewer crash fix

takahito-tejima:
two bugs fixed.

- use patchtables instead of patchTables (yes, quite confusing...) patchtables may be created ondemand if patchTables isn't given as an argument.

- If there's a hole, limit stencils may not be found for the given limit location, resulting invalid stencil entries. Using numLimitStencils to copy resulting stencils fixes that bug.


davidgyu:
We should fix the patchtables vs patchTables and cvstencils vs cvStencils confusion. The distinction is much to subtle (and brittle) for code that multiple people have to maintain.

PR #512, opened 2015-05-21
remove maximum valence limitation in gregory basis and bspline basis end caps.

takahito-tejima:
Add a non table-based version of end cap tangent computation in gregory basis, and remove MAX_ELEM and MAX_VALENCE limitations.

Also add high-valence regression shapes into glViewer.

note 1: LegacyGregory patch drawing is still under the limitation of maximum valence 29, mostly due to GLSL interstage parameter data width.
note 2: Creating end cap stencils for high-valence shape can be very slow.


takahito-tejima:
just to be clear, other valence limitation and inefficiency of LegacyGregory patch drawing are
- currently we encode quadoffsets into 8-bit
- we pack a vertex valence buffer to be the same stride with max valence for all vertices.

PR #513, opened 2015-05-22
implements Osd::CpuEvaluator::EvalPatches function.

takahito-tejima:
- it takes number and pointer for the input PatchCoords.
- add derivative evaluations.
- enhance glEvalLimit example to see the derivative evaluation works.


davidgyu:
Excellent!

PR #514, opened 2015-05-22
Replacing obsolete TopologyRefiner methods with TopologyLevel

barfowl:
This is a first pass to getting rid of the TopologyRefiners many level-specific queries outside the Far layer itself.  Once all usage everywhere has been replaced, I'll remove the unused methods from TopologyRefiner.


barfowl:
Okay, this should be good to go now.  Travis CI helped catch a few things I'm not building locally right now.  All of those level-specific methods have now been removed from far/topologyRefiner.h.

PR #515, opened 2015-05-22
358

ielillo:
Related to Dev #462. Merged dev branch into my own branch to test the changes. Had to modify the several files to make the merge successful.
First of all the functions that used to be in glCommon.cpp/h were moved to glUtils.cpp/h since the file dissappeared. As a consequence I had to search for all the functions occurrences defined in that file and make sure that were being called with the right namespace. 
The function return glewIsSupported("GL_ARB_tessellation_shader") was changed to 
glewGetExtension("GL_ARB_tessellation_shader") in the function functionSupportsAdaptiveTessellationsince since in my particular case the function would return true even though it is not implemented for my particular graphic because glew sees that the function is defined, but not that there is a real implementation .
Modified the glViewer.cpp file to merge the changes. That meant deleting functionality that was not needed, figure out merging issues, especially in the include section to include the right files to make the file compile plus make sure the that the right shader source was being used.
Finally had to add define CL_USE_DEPRECATED_OPENCL_2_0_APIS macro to suppress some warning related to deprecated functions in opencl.


jcowles:
Thanks so much for untangling the merge! As you can tell, we're really actively preparing for 3.0 release, so while it was painful, I think this is going to make it in time to be included :)

Couple notes:

 * New function names don't follow convention, for example "SupportsAdaptiveTessellation()" vs the new method "get_major_minor_version()"
 * In uses_tesselation_shaders(), are the bools intended to be static?
 * I wonder if these outputs even need to be cached, was the intent of making them static a performance concern?

Thanks again for taking the time to sort this out!

ielillo:
Ok, let me refactor the function names to use the project conventions.
Now related to the uses_tesselation_shader I'll merge it with the one above with the same name. 
Yes I wanted to cache them since I read on the glew documentation that the query functions might be slow and since the capabilities don't change during execution time you can use the cached value.

takahito-tejima:
Thanks Ignacio, I merged your changes. However I had to comment out glewGetExtension and stick to glewIsSupported since glewGetExtension doesn't work if the context is core profile (glGetString(GL_EXTENSIONS) is deprecated).
Is there any other workarounds?

PR #519, opened 2015-05-22
Faster, simpler stencil table construction

jcowles:
This is a new implementation of the stencil table construction algorithm found
in protoStencil.h. In local tests with production assets, the new algorithm is
~25% faster and significantly more stable, in terms of average performance

In one asset test, generating stencils for level 10 adaptive refinement of
BuzzLightyear was reduced from 18s to 13s.

 * Stencil construction logic is moved into ::Internal namespace
 * Removed "ProtoStencils"
 * Simplified stencil construction logic
 * Broke some unnecessary dependencies on proto stencils

This change does not change functionality or public API.


barfowl:
I'm getting some new warnings with ICC in my local builds from
far/stencilTable.h in the body of the anonymous helper function
copyStencilData():

    remark #1599: declaration hides variable "off" (declared at line 60)

I was going to simply correct them but then realized the context of this
function.

Is there any need for copyStencilData() to be declared in the header file?
Both constructors that call it can be non-inline and copyStencilData()
moved to stencilTable.cpp.  There are no template parameters or any reason
for this or anything that uses to be inline that I can see.

jcowles:
Nope, actually I was noticing that about that file in general too -- there is no cpp file currently, but it should have one and we should shift some of the contents there.

Will fix tonight!

--
Jeremy

jcowles:
Fixed and filed a new pull-request.

PR #521, opened 2015-05-23
Moved primvar interpolation methods to new Far::PrimvarRefiner class

barfowl:
This change moves all generic primvar related methods (involving template parameters <T> et al) from Far::TopologyRefiner to the new public Far::PrimvarRefiner class.  Interfaces for the <T,U> classes remain unchanged.  All usage in tutorials and examples was updated.


barfowl:
Now that this is clean I'll consolidate into a single commit, close this and resubmit.

c64kernal:
Ha we collided.

barfowl:
Nice catch on the protected/public: -- no, that's not intentional.  A local copy I tweak in my repository for some additional testing that should have been removed.  I'm reassembling the new commit and pull request now (probably an easier way if I wasn't such a git newbie...)

c64kernal:
No worries!  For my part at least, I don't mind the list of commits you had at all.

PR #522, opened 2015-05-24
Moved primvar interpolation methods to new Far::PrimvarRefiner class

barfowl:
This change moves all generic primvar related methods (involving template parameters et al) from Far::TopologyRefiner to the new public Far::PrimvarRefiner class. Interfaces for the classes remain unchanged. All usage in tutorials and examples was updated.

The previous commits in the previous pull request have been consolidated.


jcowles:
Would it be appropriate for the refiner to provide an accessor to a primvar interpolator instead?

That is, instead of this:
Far::PrimvarRefiner(*refiner).InterpolateFaceVarying(values, values + nCoarseValues);

What about this:
refiner->**GetPrimvarRefiner()**.InterpolateFaceVarying(values, values + nCoarseValues);

Would this also fix the lifetime concerns, similar to level?

c64kernal:
There's something I like about the dependencies only going from primvar refiner to topology refiner.  With the new proposal they would be interdependent.

jcowles:
Ah, yeah. 

This is true for Far::TopolyLevel as well... do you have any ideas to bring that same clarity of lifetime without the circular ref?

jcowles:
Also, if we envision a future with weak/shared ptrs in OpenSubdiv, this design would be obviated.

--
Jeremy

Sent from my phone

c64kernal:
Ah, good point about Far::TopologyLevel.  I think you're right that it would be nice for the patterns to stay consistent (though for some reason I see TopologyLevel as more intertwined with TopologyRefiner by its definition).  Barry do you have a preference?

barfowl:
I see the TopologyRefiner as one of the key building blocks in Far and the PrimvarRefiner as just another class that uses it to achieve its work -- much like the PatchTable or StencilTable factories require it to do what they need.  So I definitely don't want to see the TopologyRefiner dependent on the PrimvarRefiner.

So I don't see the ToplogyLevel and PrimvarRefiner in the same context.  The TopologyLevel is a component of the TopologyRefiner while the PrimvarRefiner is a client of it.  We return references to TopologyLevels just as we return references to internal std::vectors of other classes -- their construction and lifetime is purely the business of the class that owns them.

The PrimvarRefiner is like the PatchTable or Stencil table that are optional and are created independently but use the TopologyRefiner for their lifetime.  That's the pattern I think we should probably be following.

jcowles:
Yeah, agreed.

PR #525, opened 2015-05-26
Moved Vtr classes within internal namespace:

barfowl:
This changes set moves all major Vtr classes into the namespace internal -- discouraging them from public use (given we must export for internal usage).  All references to Vtr classes within Far was appropriately updated.  There didn't appear to be any external usage of Vtr.


c64kernal:
Sounds good, thanks Barry.  I'll change the capitalized one in Far.

PR #533, opened 2015-05-27
Lower-case Internal namespace to match convention.

c64kernal:
Renamed "Internal" namespace to "internal" to match the convention in the rest
of the library.


jcowles:
I had this queued up locally too, thanks!

--
Jeremy

PR #551, opened 2015-05-29
add c++11 option for ICC

shengfuintel:
ICC need the option "-std=c++11" to support c++11 features. The build has been tested with ICC15.0.3.


jcowles:
Thanks for the pull requst! 
But I'm curious, why does icc need C++11 enabled? If we're inadvertently using C++11 features, we should fix the offending code.

jcowles:
Looks like Takahito merged in my other change, please let us know if it's still not working on icc!

shengfuintel:
Yes, your merge fixed ICC compile error. Thanks!

PR #552, opened 2015-05-29
Fix GLEW for Core profile, minor cleanup

jcowles:
We will need to find a solution for Ignacio's older setup, but this at least
fixes the other fallout.

 * Use glewIsSupported instead of glewGetExtension
 * Convert tabs to spaces
 * Remove GLEW static caching (they only get called once)
 * Fix non-const shader string warnings introduced with GL refactoring

### DIFF
--- a/examples/common/glFramebuffer.cpp
+++ b/examples/common/glFramebuffer.cpp
@@ -143,7 +143,7 @@ GLFrameBuffer::compileProgram(char const * src, char const * defines) {
 
     GLuint program = glCreateProgram();
 
-	const std::string &versionStr = GLUtils::GetShaderVersionInclude();
+    const std::string &versionStr = GLUtils::GetShaderVersionInclude();
     static char const vtxDefineStr[] = "#define IMAGE_VERTEX_SHADER\n",
                       fragDefineStr[] = "#define IMAGE_FRAGMENT_SHADER\n";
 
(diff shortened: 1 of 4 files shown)


jcowles:
@ielillo can you try this on your machine? If it's still an issue, would you mind trying to come up with a run-time switch?

The previous change broke on modern hardware, so we need to revert the use of glewGetExtension.

ielillo:
There's I've tried running it on my machine but it still gave shader compilation issues, however if I disable the glewExperimental flag, the shader compilation works so I think I can make a runtime switch to enable or disable that flag, what do you think?

jcowles:
Wait, were already doing this, no?

Line 1879 in glViewer.cpp.

--
Jeremy

ielillo:
As a compilation flag, not as a runtime switch flag. What I'm saying is to create a runtime flag to enable or disable the experimental features.

ielillo:
Just to update everyone, I pushed to my local branch the possibility of
disabling the glewExperimental flag based on a flag given to cmake which
sets a compiler flag to disable it.

2015-05-31 0:07 GMT-03:00 Jeremy Cowles <notifications@github.com>:

> Wait, were already doing this, no?
>
> Line 1879 in glViewer.cpp.
>
> --
> Jeremy
>
> Sent from my phone
>

jcowles:
So disabling experimental fixes your issue? 

Also, a runtime solution (as opposed to a compile time flag) would be ideal, since we want OpenSubdiv to work as a syslib that the user doesn't have to compile...

--
Jeremy

ielillo:
yes, disabling it fixes my issue.

ielillo:
Just to let you know I've added a runtime flag to disable the
glewExperimental flag.

2015-06-02 23:58 GMT-03:00 Ignacio Lillo Razeto <
ignacio.lillo.razeto@gmail.com>:

> yes, disabling it fixes my issue.
>
>
> El martes, 2 de junio de 2015, Jeremy Cowles <notifications@github.com>
> escribió:
>
>> So disabling experimental fixes your issue?
>>
>> Also, a runtime solution (as opposed to a compile time flag) would be
>> ideal, since we want OpenSubdiv to work as a syslib that the user doesn't
>> have to compile...
>>
>> --
>> Jeremy
>>
>> Sent from my phone
>>

jcowles:
Cool! Want to send a new pull request?

I took a peek at it, would you also mind changing the flag to "-x"?

ielillo:
Sure, I'll do a new pull request and set the flag to -x.

2015-06-05 21:08 GMT-03:00 Jeremy Cowles <notifications@github.com>:

PR #554, opened 2015-05-30
Minor improvements to far/tutorials

barfowl:
Aside from rewriting far/tutorial_2 for a different take on varying data (as noted in #548) this change set cleans up some minor inefficiencies I've noted in the past.  It also replaces uses of the multi-level Interpolate() methods, which will soon be made obsolete.

PR #555, opened 2015-05-30
Removed multi-level interpolation methods of Far::PrimvarRefiner

barfowl:
All Interpolate() methods of Far::PrimvarRefiner that iterated over the entire hierarchy have now been removed.  These were the ones taking T* and U* that would not work correctly with user-defined types that had overloaded [].  All remaining methods should now make consistent use of T&.

Some of the documentation regarding <T,U> may warrant closer inspection, as well as renaming the template parameters more meaningfully.


barfowl:
Looks like I somehow got a stale version of a file in here somehow...  hang on...

barfowl:
Of course -- I need to have the far/tutorials pulled from #554 before a build will succeed here.

I'll close for now, will consolidate the change and submit again once the revised tutorials have been pulled.

PR #556, opened 2015-05-30
Removed multi-level interpolation methods of Far::PrimvarRefiner

barfowl:
All Interpolate() methods of Far::PrimvarRefiner that iterated over the entire hierarchy have now been removed. These were the ones taking T* and U* that would not work correctly with user-defined types that had overloaded []. All remaining methods should now make consistent use of T&.

Some of the documentation regarding may warrant closer inspection, as well as renaming the template parameters more meaningfully.

Additional minor improvements were made to some of the far/tutorials while replacing the old methods.  far/tutorial_2 in particular was rewritten with a different use case for varying data.

### DIFF
--- a/examples/farViewer/farViewer.cpp
+++ b/examples/farViewer/farViewer.cpp
@@ -649,7 +649,14 @@ createFarGLMesh(Shape * shape, int maxlevel) {
                 float const * ptr = &shape->uvs[i*2];
                 values[i].SetPosition(ptr[0],  ptr[1], 0.0f);
             }
-            Far::PrimvarRefiner(*refiner).InterpolateFaceVarying(values, values + nCoarseValues);
+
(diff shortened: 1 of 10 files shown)


barfowl:
This request combined #554 and #555 that I cancelled.  All looks good with all of the changes together now.

jcowles:
Cool -- looking at how common the pattern is, I'm having second thoughts about not supporting a single function to interpolate all levels, but I'll merge this in and then we can see if we can do better by Monday (seems unlikely).

barfowl:
Yeah, I regretted removing this.  I wouldn't say I'm having second thoughts
(it does remove the T* vs T& confusion), but hopefully we can get something
back in.  I have an idea that may be viable that I'll run by you shortly.

PR #557, opened 2015-05-30
OpenGL init without errors, example cleanup

jcowles:
The GLFW context version hint is a minimum version, not maximum version so
requesting 4.4 and then falling back to lower versions doesn't make sense.

This change sets the minimum version to 3.2 and attempts to standardize this
across all example apps.

Also print the maximum supported GL version along with the context version
at startup.

### DIFF
--- a/examples/common/glUtils.cpp
+++ b/examples/common/glUtils.cpp
@@ -26,6 +26,8 @@
 #include <string>
 #include "glUtils.h"
 
+#include <GLFW/glfw3.h>
+
 namespace GLUtils {
 
 // Note that glewIsSupported is required here for Core profile, glewGetExtension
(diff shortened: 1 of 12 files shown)


takahito-tejima:
So... with this change merged, I started getting errors on example viewers, because the GL context is initialized 3.2. (not higher).

ubuntu 14.02 / glfw 3.0.2 / nvidia 343.19  : Init OpenGL 3.2
rhel 6.5 / glfw 3.0.4 / nvidia 340.46 : Init OpenGL 3.2
OSX 10.10.3 / glfw 3.1.0 / GeForce (310.41.25f01) : Init OpenGL 4.1
OSX 10.10.3 / glfw 3.1.0 / Intel HD Graphics 4000 (10.6.20) : Init OpenGL 4.1

    // Here 3.2 is the minimum GL version supported, GLFW will allocate
    // a higher version if possible.
    int major = 3,
        minor = 2;
is this true? or maybe I need to use the newer version of GLFW?

jcowles:
That's the claim in the docs and it indeed works that way on OSX... 

--
Jeremy

takahito-tejima:
yeah, it works on OSX for me too but not on linux (GLX). And I updated GLFW to 3.1.2 but I'm still getting errors. In glfw/src/glx_context.c, 

        // NOTE: Only request an explicitly versioned context when necessary, as
        //       explicitly requesting version 1.0 does not always return the
        //       highest version supported by the driver
        if (ctxconfig->major != 1 || ctxconfig->minor != 0)
        {
            setGLXattrib(GLX_CONTEXT_MAJOR_VERSION_ARB, ctxconfig->major);
            setGLXattrib(GLX_CONTEXT_MINOR_VERSION_ARB, ctxconfig->minor);
        }

So it looks like we shouldn't set versions except OSX...?

jcowles:
I'll try setting no version on OS X and see if that works too.

I realized that the docs I found leave it ambiguous, they say it's the minimum version, but it doesn't say that a greater version will be used when available.

--
Jeremy

jcowles:
Not setting the minimum version does not work on OSX.
So perhaps we should do as you suggested and only set the version on OSX.

jcowles:
I did some digging and I think I got to the bottom of it -- we should explicitly do this on OSX or perhaps whenever ARB_compatibility is not available.

Here is the documentation I was referring to that says it's a "minimum" version hint:
http://www.glfw.org/faq.html#how-do-i-create-an-opengl-30-context

And I just noticed it *is* under the "OSX" section, which I guess is implying its different on other operating systems. 

I think this may explain why it's different (specifically the last two sentences):

> 1.5 - What versions of OpenGL are supported by GLFW?
> 
> In short, all versions provided by your operating system.
> 
> This question likely stems from the original version of the WGL_ARB_create_context and 
> GLX_ARB_create_context extensions, which stated that the old context creation mechanism would 
> be limited to OpenGL version 2.1. However, this is no longer the case and the specifications have 
> been updated accordingly.
> 
> GLFW 2.6 and earlier use only the older context creation mechanism, which on Windows and X11 
> may return contexts of any version, however new, provided they implement the ARB_compatibility 
> extension. Most modern drivers do this.
> 
> Explicit creation of OpenGL contexts of version 3.0 and above on Windows and X11, including 
> profiles and flags, is supported by GLFW 2.7 and later.
> 
> However, OS X did not support OpenGL 3.0 or later at the time that GLFW 2.7 was released, and 
> the support that Apple has since added only includes forward-compatible OpenGL 3.2 core profile 
> contexts.

c64kernal:
Thanks Jeremy.  Are you planning on doing this?

jcowles:
I just made a new pull request!

c64kernal:
Just saw, looks great, will merge in as soon as it goes green and will try it on both Linux and Windows (both of which currently fail).  Thanks again!

PR #559, opened 2015-05-30
Re-organized patch tessellation shader code

davidgyu:
This change refactors the GLSL and HLSL patch shader code so that
most of the work is implemented within a library of common functions
and the remaining shader snippets just manage plumbing.

There is more to do here:
  - varying and face-varying data can be managed entirely by the client
  - similarly, displacement can be implemented in client code
  - there's still quite a bit of residual boiler-plate code needed
    in each shader stage that we should be able to wrap up in a more
    convenient form.


takahito-tejima:
I locally merged this change and looked through. This is an awesome cleanup, thank you!

jcowles:
Looks like you got a bum .swp file in here...

jcowles:
*vim* swp file, that is

PR #565, opened 2015-05-31
Only specify an OpenGL version hint on OSX

jcowles:
It appears the behavior of the version hint is operating system
(and perhaps driver) specific. On OSX the version hint is a
minimum version, but on Linux you get exactly the version
requested.

The GLFW FAQ seems to suggest that any driver not supporting
ARB_compatibility may need the hint.

s

### DIFF
--- a/examples/common/glUtils.cpp
+++ b/examples/common/glUtils.cpp
@@ -39,11 +39,6 @@ namespace GLUtils {
 
 void
 SetMinimumGLVersion() {
-    // Here 3.2 is the minimum GL version supported, GLFW will allocate
-    // a higher version if possible.
-    int major = 3,
-        minor = 2;
-
     #define glfwOpenWindowHint glfwWindowHint
     #define GLFW_OPENGL_VERSION_MAJOR GLFW_CONTEXT_VERSION_MAJOR
     #define GLFW_OPENGL_VERSION_MINOR GLFW_CONTEXT_VERSION_MINOR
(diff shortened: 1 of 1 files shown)


c64kernal:
Sadness.  Both Linux and Windows:

GLFW Error (65540) : Context profiles only exist for OpenGL version 3.2 and above
Failed to create OpenGL context.

takahito-tejima:
I think if we don't specify versions (on linux and windows), CORE_PROFILE shouldn't be set.
So put CORE_PROFILE hint into __APPLE__ would work...

c64kernal:
Thanks Takahito!  Just trying that!

c64kernal:
Beautiful.  Pull request incoming.

jcowles:
Sweet.

jcowles:
And thanks for testing, George!

c64kernal:
Thank you guys!  Fix pull requested.

takahito-tejima:
Thanks for testing. But then we lose an ability to run with pure core profile context on linux and windows. So far I only needed that to profile with NSight which only supports core profile program.
It's not so important and we can easily tweak if we really need, so I think not specifying core profile on linux and windows sounds ok.

c64kernal:
Takahito, it feels like we should have an explicit run-time switch to force a particular gl context (profile, version).  That will be useful for testing generally.

takahito-tejima:
That is awesome!

PR #567, opened 2015-05-31
Fix GL context init for Linux and Windows.

c64kernal:
Move the core profile specification into the OSX only section.


jcowles:
I just noticed, FORWARD_COMPAT should be in CORE_PROFILE as well -- mind updating?

jcowles:
That is, glfwOpenWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GL_TRUE); should be inside the #ifdef CORE_PROFILE guard.

c64kernal:
Done!

PR #568, opened 2015-05-31
Command line options for GL context.

c64kernal:
Adding command line options to glViewer to make it easy
to control the requested GL version and profile.  While
it is only enabled for glViewer in this change, it will
be easy to extend to all our example viewers.  The new
command line options are:

    -glCoreProfile on|off
    -glForwardCompat on|off
    -glVersion M.n


takahito-tejima:
This is great! thank you very much.

PR #576, opened 2015-06-01
Removed unnecessary files included by public headers in Far

barfowl:
This set of changes was initially to remove a block of include files from far/TopologyRefiner.h that are now no longer necessary.  I looked around at a few other public headers and found a few more that appeared to be unnecessary and removed them as well.


barfowl:
Closing to consolidate and resubmit.

PR #578, opened 2015-06-01
Added include of algorithm for min/max

davidgyu:


barfowl:
Is that the result of me pulling it out of patchTable.h in #577?  How has
it managed to pass so many builds until now?

davidgyu:
I'm not sure. I think this is just the usual indeterminacy of platform headers. I hit this right off with my VS 2013 build. The Linux and Mac builds are fine with or without.

PR #580, opened 2015-06-01
Renaming of face-varying methods to access values per face

barfowl:
As discussed, I replaced the string FVarFaceValues with FaceFVarValues in a couple of methods in Far -- accessing FVarValues through the TopologyLevel and assigning the base mesh in the RefinerFactory.  I also update the default factory's assignment of face-varying data to be more consistent with the desired terminology.

There is some internal usage in Vtr that still retains the old FVarFaceValue term that I'll deal with soon with some other minor changes to Vtr.


davidgyu:
I think this helps a lot. It still would be nice to have some more documentation/comments for these methods. Thanks! -David

barfowl:
Coming soon.  I had a look at far/tutorial_3 and while it thankfully didn't further confuse things, it wasn't particularly helpful either.  If you get a chance to have a look at it and have any ideas, let me know.

PR #582, opened 2015-06-02
Disabling feature-adaptive refinement for non-Catmark schemes

barfowl:
While the feature-adaptive code path is now far more supportive of Loop, its correctness is not assured and there are definitely downstream limitations when it comes to patches.  So until its been more rigorously approved, we're disabling feature-adaptive for anything other than Catmark (which is the state of 2.x).


c64kernal:
Thanks very much, Barry.  It'd be great if this found itself somewhere in the documentation too when we do the doc pass.

barfowl:
I'll make a note to at least put something in the Doxygen tags in the
header for the RefineAdaptive() method.

PR #584, opened 2015-06-02
Unlabel rotation of single crease patch.

takahito-tejima:
Use boundaryMask to identify the crease edge from 4 edges, instead of permuting index upfront.
This fix gives consistent patch coordinate on single-crease patches.

In shader, experimentally use the same infinite sharp matrix for both boundary and single-crease
patch. There's still a room for optimization in shader code though.

### DIFF
--- a/examples/dxViewer/dxviewer.cpp
+++ b/examples/dxViewer/dxviewer.cpp
@@ -91,10 +91,16 @@ enum KernelType { kCPU           = 0,
                   kCL            = 4,
                   kDirectCompute = 5 };
 
-enum DisplayStyle { kWire = 0,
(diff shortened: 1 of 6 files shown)


barfowl:
This issue reminded me of a consideration for the future...  With
triangular patches the edge boundary mask alone is not enough to identify
all we need for a boundary patch, i.e. unlike quad patches triangular
patches can have a vertex on the boundary but no boundary edges.  So we may
eventually need both edge- and vertex-boundary masks, or something that
indicates the boundary mask refers to vertices rather than edges.  At this
point I don't think this is publicly exposed, so we may be free to do what
we need.

PR #586, opened 2015-06-02
Renamed per-patch face-varying access methods

davidgyu:
This matches more closely the pattern used elsewhere in Far:
    GetFVarPatchValues() --> GetPatchFVarValues()
    GetFVarPatchesValues() --> GetFVarValues()

### DIFF
--- a/examples/farViewer/farViewer.cpp
+++ b/examples/farViewer/farViewer.cpp
@@ -393,7 +393,7 @@ createFVarPatchNumbers(OpenSubdiv::Far::PatchTable const & patchTable,
         handle.patchIndex = patch;
 
         OpenSubdiv::Far::ConstIndexArray const cvs =
-            patchTable.GetFVarPatchValues(channel, handle);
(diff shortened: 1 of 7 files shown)


barfowl:
Is this all going to be hidden in 3.0 with the rest of the face-varying
patch work that's in progress, or is this functional to access the bilinear
face-varying values?

davidgyu:
Right, this is part of the interface that is needed to access the bilinear face-varying values. I have a couple of follow up changes which will isolate and then remove the bicubic face-varying support from Far.

barfowl:
Okay.  The trend in the existing topological face-varying queries is that
the channel argument come last, e.g.  GetFaceFVarValues(faceIndex,
channel).  The channel also usually defaults to 0 if unspecified in most
queries -- the expectation being that a lot of use cases only have one.

davidgyu:
Good point. I'll fix that too. Thanks!

PR #587, opened 2015-06-02
Prepare to defer face-varying bicubic patches

davidgyu:
There's a lot of good foundational work here to eventually support
smooth interpolation of face-varying patches. Unfortunately, this
is not quite ready to release. Therefore, we've decided to defer this
feature until a later release.

This change hides this code behind the FAR_FVAR_SMOOTH_PATCH macro.


davidgyu:
My intention is to follow this with a change which deletes all of the code marked off by the macro. So, that for the release both the code and the macro will be gone.

davidgyu:
The follow up change is here if you want to see where this is headed: https://github.com/davidgyu/OpenSubdiv/commit/2d6b9f4e2d56c44f4fdbe4f4f010643616bfe740

PR #600, opened 2015-06-04
Made channel argument non-optional for internal face-varying methods

barfowl:
The use of Class::FVarMethod(..., channel = 0) is convenient for users, but not for internal development.  We've run into a couple of places where the channel argument wasn't properly propagated to another method because of the default value, so I have removed it from internal methods.  In doing so I found one bug that we would have hit with the face-varying patches.


jcowles:
Nice catch!

davidgyu:
Great! Thanks!

Issue #604, opened 2015-06-04
RC1 : XineramaQueryScreens missing symbols


nyue:
Hi,

  I thought I'd chime in since we are on the release candidate phase.

  On CentOS 6.4, there is also a need to link with Xinerama library so it is not just Ubuntu that has the issue.

[ 35%] Building CXX object regression/osd_regression/CMakeFiles/osd_regression.dir/main.cpp.o
Linking CXX executable ../../bin/osd_regression
/corp/projects/eng/nyue/Systems/glfw/v3.1.1/lib/libglfw3.a(x11_init.c.o): In function `initExtensions':
x11_init.c:(.text+0x1af0): undefined reference to `XineramaQueryExtension'
x11_init.c:(.text+0x1b0a): undefined reference to `XineramaIsActive'
/corp/projects/eng/nyue/Systems/glfw/v3.1.1/lib/libglfw3.a(x11_monitor.c.o): In function `_glfwPlatformGetMonitors':
x11_monitor.c:(.text+0x6fc): undefined reference to `XineramaQueryScreens'
collect2: error: ld returned 1 exit status
make[2]: *** [bin/osd_regression] Error 1
make[1]: *** [regression/osd_regression/CMakeFiles/osd_regression.dir/all] Error 2
make: *** [all] Error 2

Cheers

jcowles:
This should be fixed now -- please reopen if that's not the case!

Issue #473, opened 2015-05-13
undefined reference to `XineramaQueryScreens' while building the examples


opeca64:
Hit a simple bug inside of the examples.  I believe this is Ubuntu-specific. ( I have turned off <code>CLEW</code> because of mentioned bug in #472 )

<pre><code>
MYUSERNAME@MYCOMPUER:/home/pion/ext/openSubdiv-current/build$ make
[  0%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuKernel.cpp.o
[  1%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuComputeController.cpp.o
[  1%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuComputeContext.cpp.o
[  1%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalLimitContext.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalLimitController.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalLimitKernel.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalStencilsContext.cpp.o
[  3%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalStencilsController.cpp.o
[  3%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuSmoothNormalContext.cpp.o
[  4%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuSmoothNormalController.cpp.o
[  4%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuVertexBuffer.cpp.o
[  4%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/drawContext.cpp.o
[  5%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/drawRegistry.cpp.o
[  5%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/evalLimitContext.cpp.o
[...]

PR #605, opened 2015-06-04
Added Xinerama as a link dependency for GLFW

davidgyu:
This is a recently introduced dependency in GLFW 3.1.x

Since Xinerama is widely supported, simply adding this as a required
dependency for now. If this turns out to be a problem then we can
make this dependency optional when using older versions of GLFW.


jcowles:
Should we also make this requirement conditional based on the GLFW version in which it was introduced?

This fixes #604 and #473

PR #607, opened 2015-06-05
Fixes for farViewer

davidgyu:
Previously, the farViewer was raising GL invalid operation errors when switching isolation levels.
These resulted from a combination of bad GL calls made to collect query results and to display the control mesh of the current model.

This pull request fixes these two issues and also makes a few changes to the hotkey bindings so that the farViewer is more consistent with the other example viewers.

We may want to consider taking out the performance metric code altogether, since that is not the point of this example.


davidgyu:
Closing, this was intended for 'dev'

Issue #472, opened 2015-05-13
nvcc fatal: Value 'compute_11' is not defined for option 'gpu-architecture' & ccmake not finding 'clew'


opeca64:
Hey!

It appears that the --gpu-architecture flag for <code>nvcc</code> was hard-coded in <code>$(ROOT)/opensubdiv/CMakeLists.txt </code> at line 109...unfortunately it no longer compiles as written against the new Cuda 7 Toolkit on Ubuntu Linux...

DOESN'T COMPILE:
<code>
            list( APPEND CUDA_NVCC_FLAGS -Xcompiler -fPIC --gpu-architecture compute_11 )
</code>

COMPILES:
<code>
            list( APPEND CUDA_NVCC_FLAGS -Xcompiler -fPIC --gpu-architecture compute_20 )
</code>

Would it  be worth getting the proper arch automatically/dynamically using the built-in MACROS ( __CUDA_ARCH__ etc. )?  I'm not familiar enough with the OpenSubdiv code base to answer that...  Should this be changed via a pull-request or will I just have to manually fix it if the cmake files get altered later down the line? 

Here's the full ccmake compiling error:
<pre><code>
MYUSERNAME@MYCOMPUTERNAME:/home/pion/sandbox/openSubdiv-current/build$ make
Scanning dependencies of target sdc_obj
[  1%] Building CXX object opensubdiv/sdc/CMakeFiles/sdc_obj.dir/crease.cpp.o
[  1%] Building CXX object opensubdiv/sdc/CMakeFiles/sdc_obj.dir/typeTraits.cpp.o
[  1%] Built target sdc_obj
Scanning dependencies of target osd_cpu_obj
[  1%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuKernel.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuComputeController.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuComputeContext.cpp.o
[  2%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalLimitContext.cpp.o
[  3%] Building CXX object opensubdiv/osd/CMakeFiles/osd_cpu_obj.dir/cpuEvalLimitController.cpp.o
[...]

PR #617, opened 2015-06-09
Fix for Bug #472, nvcc fatal compute_11 unsupported workaround. Better cmake hints for Ubuntu

opeca64:
- Added proper hints for TBB and CLEW in the cmake find-files to reflect their respective default and likely installation paths on Ubuntu14.04.1
- Corrected 'cg' and 'cgGl' to 'Cg' and 'CgGl' to reflect the default Autodesk naming scheme for the Cg libraries in Maya 2015 x64 for Linux
- Simply changed 'compute_11' to 'compute_20', for a workaround regarding the fatal crash for nvcc.  Detection should be implemented.

### DIFF
--- a/opensubdiv/CMakeLists.txt
+++ b/opensubdiv/CMakeLists.txt
@@ -106,7 +106,7 @@ if (NOT NO_LIB)
     if( CUDA_FOUND )
         include_directories( "${CUDA_INCLUDE_DIRS}" )
         if (UNIX)
-            list( APPEND CUDA_NVCC_FLAGS -Xcompiler -fPIC --gpu-architecture compute_11 )
+            list( APPEND CUDA_NVCC_FLAGS -Xcompiler -fPIC --gpu-architecture compute_20 )
         endif()
     endif()
 


jcowles:
Hi Lindsey, have you signed a CLA (http://graphics.pixar.com/opensubdiv/docs/contributing.html) for us yet?

Sorry to let this linger, we've been debating internally the best way to handle this -- as you indicated in your commit comment, it seems like this is a lateral move, but the ultimate right solution requires more digging.

Thanks!

opeca64:
Hi Jeremy, thanks for catching up.  Since I opened up Bug #472, I haven't had a lot of time to dig into OpenSubdiv, and I felt it odd for me to make a dynamic fix for it because I have only recently started using OpenSubdiv.  However, there's a good chunk of time between now and SIGGRAPH 2015 in which I'll be happy to attempt an actual solution for the community.  

Is anyone from the team going to around for SIGGRAPH?  If so, we can talk more about it then.  

Pixar Human Resources should have all of my contact information because they had contacted me regarding an internship this past winter.  I'll review the CLA.

jcowles:
Yes, we'll be at SIGGRAPH and it would be great to meet up!

And to be clear, I'm not suggesting that you need to take this on, I'm just explaining why we've been hesitant to merge this request.

opeca64:
Great!  Do you know if there's going to be a Birds of a Feather Session this year for OpenSubdiv?  

If there turns out to be no internet at the conference ( this has happened ), Beth Sasseen, in HR, has my cell number.  Feel free to text/email/call me if we haven't yet connected at the BOF or elsewhere.

If I can see a problem, I know can solve it, and I should fix it especially if no one else can get to it - that's what makes me an R&D software developer. Hence why I feel swamped all the time. ;)

davidgyu:
We addressed the main underlying issue with PR #869 which both sets a better fallback GPU architecture for recent CUDA releases and exposes a way to specify additional CUDA build options.

The OpenCL compatibility issues will be addressed by moving to the standard FindOpenCL module as described in the comments for #913

Thanks!

PR #620, opened 2015-06-10
Fix a bug in limit stencil table creation

takahito-tejima:
LimitStencilTableFactory::Create was returning fewer number of
stencils, off by the number of coarse vertices.


jcowles:
This looks good to me, just waiting for the build to come back green.

PR #623, opened 2015-06-10
More revisions to documentation pages

barfowl:
I updated the Roadmap page as discussed to reflect more recent activity and interest.

I also re-ordered some of the topics in the Subdivision Surfaces page to de-emphasize non-manifold topology and some of the other less common features -- given that this is intended as an intro to the topic.  With the non-manifold restrictions now removed, its much less important.

Finally, I re-ordered some of the topics in the Release Notes and did some grouping that warrants review.  Most of the topics seemed to be either Performance or Topology related, so I grouped them as such -- which mainly involved items under Topology since Performance was already evolving that way.  I tried to indent each item so that the item titles stand out more, but had some issues with the RST formatting that I couldn't resolve -- indented most as expected, but a few strange exceptions.

### DIFF
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -51,109 +51,145 @@ algorithms that greatly enhance performance over previous versions.
 
 **Faster Subdivision**
 
-A major focus of the 3.0 release is performance. It should provide
-"out-of-the-box" speed-ups close to an order of magnitude for topology
-refinement and analysis (both uniform and adaptive).
+ A major focus of the 3.0 release is performance. It should provide
+ "out-of-the-box" speed-ups close to an order of magnitude for topology refinement
+ and analysis (both uniform and adaptive).
+
+**Introducing Stencil Tables**
+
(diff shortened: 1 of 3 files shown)


jcowles:
The leading space/indentation turned those paragraphs into block quotes, was that intentional?

barfowl:
Yes, as I noted in my original comment, I was trying to indent these and I've seen that format used elsewhere in the docs.

jcowles:
Ah, I see (I did read your comment, but wasn't sure if this was the exact effect you were going for).

barfowl:
There is the effect I was going for and what I ended up with -- which may not be the same.  Can you verify the resulting HTML from this?  In my local build of the doc pages I see the first subsection of Improved Performance indented less than the rest, and the first two subsections of Topological Flexibility indented less than the following two.

barfowl:
If anyone is opposed to this new indenting, or has any alternatives on how to make the topics stand out more, let me know or go ahead and make the changes.  I found the headings and topics getting buried in all of that left-aligned text and thought this would help.

Issue #648, opened 2015-06-19
Compile error on iOS with Xcode 7


meshula:
This line won't compile (two places in fvarRefinement.cpp):
                cVertEdgeBuffer[i] = _refinement.getEdgeChildEdges(pVertEdges[i])[pVertInEdge[i]];

I hacked around it by adding a setAtIndex method to StackBuffer.
                cVertEdgeBuffer.setAtIndex(i, _refinement.getEdgeChildEdges(pVertEdges[i])[pVertInEdge[i]]);

The error is pasted below. I assume it has something to do with sizeof int making overloads ambiguous, but I'm not sure. Posting here in case anyone has ideas.

extern/opensubdiv/vtr/fvarRefinement.cpp:592:32: error: use of overloaded operator '[]' is ambiguous (with operand types 'internal::StackBuffer<Index, 16>' and 'int')
                cVertEdgeBuffer[i] = a;
                ~~~~~~~~~~~~~~~^~
In file included from 
extern/opensubdiv/vtr/fvarRefinement.cpp:27:
extern/opensubdiv/vtr/../vtr/stackBuffer.h:62:18: note: candidate function
    TYPE &       operator[](size_type index)        { return _data[index]; }
                 ^
extern/opensubdiv/vtr/../vtr/stackBuffer.h:63:18: note: candidate function
    TYPE const & operator[](size_type index) const  { return _data[index]; }
                 ^

extern/opensubdiv/vtr/fvarRefinement.cpp:592:32: note: built-in candidate operator[](int *, int)
                cVertEdgeBuffer[i] = a;
                               ^
extern/opensubdiv/vtr/fvarRefinement.cpp:592:32: note: built-in candidate operator[](const int *, int)
extern/opensubdiv/vtr/fvarRefinement.cpp:592:32: note: built-in candidate operator[](volatile int *, int)
extern/opensubdiv/vtr/fvarRefinement.cpp:592:32: note: built-in candidate operator[](const volatile int *, int)
extern/opensubdiv/vtr/fvarRefinement.cpp:663:28: error: use of overloaded operator '[]' is ambiguous (with operand types 'internal::StackBuffer<Index, 16>' and 'int')
            cVertEdgeBuffer[i] = _refinement.getEdgeChildEdges(pVertEdges[i])[pVertInEdge[i]];
            ~~~~~~~~~~~~~~~^~
In file included from 
extern/opensubdiv/vtr/fvarRefinement.cpp:27:
[...]

PR #651, opened 2015-06-19
Code tweak to prevent build failure in iOS with XCode7

barfowl:
According to issue #648 (closed as it related to another problem) a missing space between literal strings is causing a compilation error with XCode7 on iOS.  This should be a harmless fix.


meshula:
thanks

PR #663, opened 2015-06-20
Simplifying Porting Guide

c64kernal:
Moved big sections out of the porting guide and merged them into
the overview for Far documentation.


davidgyu:
I think this looks pretty good and it makes sense for more of this to be part of the main Far overview than just in the porting guide. I had been postponing updating the factory section to avoid duplicating the content from the porting guide, but just migrating the content looks good.

There's the sentence beginning: "Regardless of the three construction choices outlined below" remaining in the porting guide which seems out of place now.

c64kernal:
Good catch, thanks David!

I tried to interweave as much of the text from the porting guide over to the far docs as I could without duplication... a lot of it stuck!

PR #665, opened 2015-06-21
More updates to the Subdivision Compatibility page

barfowl:
More issues originally intended for RenderMan compatibility are moving into 2.x compatibility -- in this case the "bowtie vertex" leading to Hbr's "split" vertices in 2.x.  A couple of other minor tweaks included.


barfowl:
I realized after submitting this that the diagram didn't actually show the 3.0 result, so I'll be revising the diagram and caption.

PR #670, opened 2015-06-22
Minor overview doc changes and nav title update

jcowles:
Having two introduction sections in the navigation bar was awkward, here I've
left "introduction" as-is but changed "Introduction to 3.0" to "Release 3.0" and
"Release 3.0" to "Overview".

Also some very minor chagnes to the body of intro_30.rst

### DIFF
--- a/documentation/intro_30.rst
+++ b/documentation/intro_30.rst
@@ -22,8 +22,8 @@
      language governing permissions and limitations under the Apache License.
 
 
-Introduction to the 3.0 Release
--------------------------------
+Overview of Release 3.0
+-----------------------
 
 .. contents::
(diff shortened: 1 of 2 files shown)


barfowl:
All good changes here thanks.  One that I have second thoughts about is the parenthetical remark removed from Sdc.  I have been trying to create reasonable expectations around Sdc:  it is not going to be easy for someone to write their own feature-complete and compliant version of subdivision with Sdc.  I hope users realize that up front before investing too much time coming to that realization on their own.  Maybe more in the sdc_overview is warranted. 

I'll leave it up to you if you want to add something back here, but that's been the intention.

PR #679, opened 2015-06-23
Updated subdivision surface intro page

davidgyu:
Moved the description of uniform and adaptive refinement to
earlier in the page and also updated the table comparing the
two methods.

Previously, the table contained many restrictions related to
our use of Hbr which no longer need to be considered.

### DIFF
--- a/documentation/subdivision_surfaces.rst
+++ b/documentation/subdivision_surfaces.rst
@@ -86,6 +86,75 @@ connectivity.
 
 ----
 
+Uniform Subdivision
+===================
+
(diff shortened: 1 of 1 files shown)


barfowl:
Did we agree at one point to get rid of those Release Notes 3.0.0 note boxes?  I know we got rid of them in some contexts.

c64kernal:
We decided to make them less jarring with their red background, I believe.  Right now they don't stand out quite so much, and we decided that it was useful to have them in-context.

PR #695, opened 2015-07-05
Index out of Range Issues

dazcjones:
These changes fix some out range issue I ran into.  

The ones in "level.h" make it so on  windows debug the code does not assert and abort when there are unconnected vertices at the end of the list.

The change in "fvarLevel.cpp" fixes an index out or range when there is face varying data and unconnected vertices.

### DIFF
--- a/opensubdiv/vtr/fvarLevel.cpp
+++ b/opensubdiv/vtr/fvarLevel.cpp
@@ -411,6 +411,8 @@ FVarLevel::completeTopologyFromFaceValues(int regularBoundaryValence) {
         ConstIndexArray       vFaces  = _level.getVertexFaces(vIndex);
         ConstLocalIndexArray  vInFace = _level.getVertexFaceLocalIndices(vIndex);
 
+        if ( vFaces.empty() || vInFace.empty() )
+            continue;
         //
         //  First step is to assign the values associated with the faces by retrieving them
         //  from the faces.  If the face-varying topology around this vertex matches the vertex
--- a/opensubdiv/vtr/level.h
+++ b/opensubdiv/vtr/level.h
@@ -521,23 +521,23 @@ Level::getFaceEdges(Index faceIndex) {
 //
 inline ConstIndexArray
 Level::getVertexFaces(Index vertIndex) const {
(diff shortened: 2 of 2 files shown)


jcowles:
Thanks for the fixes! 

To accept your pull requests, we first need a signed CLA to please the gods of legality:
http://graphics.pixar.com/opensubdiv/docs/contributing.html

Is that possible?

Thanks again!

dazcjones:
DAZ3D should already have one on recorded. I should be listed on it.

jcowles:
Hrm, I actually looked previously and we don't have a record of it -- I'll track it down. 
Sorry for the confusion!

dazcjones:
It should have been sent in like back in Nov/Dec 2014. If you can't find it, I can double check it got sent in tomorrow.

jcowles:
Would you mind forwarding it along again? I confirmed it was not received previously.

dazcjones:
Sorry, it seems there was some misunderstanding internally.  I was informed that it has been sent now.

barfowl:
Thanks for catching and fixing this.  Those vectors are expected to be
non-empty, in which case taking &vector[0] with any compiler knowing that
it is empty is a bit dubious.  I may revisit this later.

The change to fvarLevel.cpp has lead me to some other oversights with
unconnected vertices and face-varying data.  I've found a couple more bugs
and will be updating them shortly.  If you do run into those in the
meantime, keep in touch.  I'll open a new issue about this shortly.

barfowl:
Do we have authorization yet to accept this request?  I've got an incremental change to these that I'd like to add to really address the issue with unconnected verts and face-varying data.

The change submitted here to far/fvarLevel.cpp does avoid the uninitialized memory read that was occurring at line 424 when accessing vFaces[0], but it then leaves vValues[0] uninitialized on the right-hand-side, which will lead to other problems later.

jcowles:
Yes, we've received the CLA, so we can accept the pull request.

It would be great to add a regression test/shape to exercise this case -- Barry, if you have a followup change and some other work in-flight, would you also mind adding the regression shape?

barfowl:
That's the plan, thanks.

PR #696, opened 2015-07-05
Fix offsets in OmpEvalStencils when start is non-zero.

dazcjones:
The start index offset was being added twice causing issues when it was non-zero.


takahito-tejima:
Thanks for the fix! we'll add a unit test to cover this bug.

PR #698, opened 2015-07-09
Fix compile warning from Intel compiler

shengfuintel:
This pull request is to clean up some compile warnings from Intel compiler.


barfowl:
I'm glad to see this getting some attention.  I use ICC and have been seeing some warnings creep back in over time.  In particular, "remark #1599: declaration hides XXX..." is definitely worth heeding -- if there is a way to get that enabled with GCC we would probably catch those sooner.

As for these changes, I'm not in favor of applying a pragma to disable a warning for an entire source file when only a few lines are relevant.  For the case of "remark #1572: floating-point equality and inequality comparisons are unreliable" we could deal with the individual cases when we need to do this, and there are only two:  comparing two sharpness values for the constant-crease case, and comparing a weight to zero.  We can deal with those locally without impacting new code that is added later that should trigger the warning.

bdcline:
GCC has a shadowing warning - it may be what you are looking for: -Wshadow .

jcowles:
I agree Barry.

Also, it would be nice to cherry-pick (or squash) the single relevant change here.

jcowles:
And thanks for cleaning up the warnings!!

shengfuintel:
Based on Barry's comments, I use #pragma warning (push) and #pragma warning (pop) to minimize the scope of the affected code. And a few minor changes to remove unnecessary semicolon. Please review it.

jcowles:
I think Barry's point was that we should fix the issues rather than suppress the warnings -- that would be my preference too.

barfowl:
Yes, if possible.  With the floating-point equality warnings the pragma declarations for each usage are pretty invasive to the code.  I'm used to having legitimate exact floating-point equality tests encapsulated in macros or inline functions -- within which the pragmas suppress the warnings.  That's what I suggest we do for this case.  We can take these local pragma declarations for now and get rid of them once we have such utilities available for usage.

Brian pointed out GCC's -Wshadow flag that would have caught of couple of the warnings fixed here, but it also seems far pickier than ICC, e.g. it warns about argument or local variable names shadowing method names in a few places.  We'd need to clean those up before enabling it (and I may start doing so regardless).

jcowles:
@shengfuintel Does my comment about squashing the changes make sense? 

Make sure you have pixar/OpenSubidv setup as "upstream":
 git remote add upstream  https://github.com/PixarAnimationStudios/OpenSubdiv
 git fetch upstream

Then if you run:
 git rebase -i upstream/dev

You can squash your changes like this (IMPORTANT: notice I moved the last commit to the top of the list):

 pick c615cb9 fix ICC compile warning
 f 9cc4a1b add c++11 option for ICC
 f e8a46c2 revert the change in CMkaeLists.txt
 f 7ef9b3c revert the change in CMkaeLists.txt
 f c138fdc Fix Intel compiler warning
 f c042980 minimize the scope of disabled warnings

jcowles:
alternatively, you could start a new branch from upstream/dev and then cherry-pick the relevant changes.

jcowles:
For example:
 git checkout -b dev-clean upstream/dev
 git cherry-pick c138fdc c042980

That might be cleaner and avoid having to push with --force.

shengfuintel:
@jcowles: I am not an expert of GIT. Can you let me know which source code change, which you do not want to take,  in shengfuintel/OpenSubdiv branch dev? I can change it, so you can simply merge them all into PixarAnimationStudios/OpenSubdiv. Thanks.

jcowles:
The source diff is fine, I just want to exclude the following changes from the commit history:

add c++11 option for ICC	 9cc4a1b
revert the change in CMkaeLists.txt	e8a46c2
revert the change in CMkaeLists.txt	7ef9b3c
Merge branch 'dev' of https://github.com/PixarAnimationStudios/OpenSu… …	a08a6a1
Merge branch 'dev' of https://github.com/PixarAnimationStudios/OpenSu… …	c2eadf8
Merge branch 'dev' of https://github.com/PixarAnimationStudios/OpenSu… …	09af02d

If you don't want to rebase or cherry-pick the relevant changes, you could start from a clean dev tree and reapply the diff with a single new commit.

shengfuintel:
@jcowles: I did the following to create a single new commit for my change:

git remote add upstream https://github.com/PixarAnimationStudios/OpenSubdiv
git fetch upstream

git checkout -b clean-icc upstream/dev

and then I merged my change into clean-icc branch.

However, I won't be able to run "git push" since I do not have the permission to push into  PixarAnimationStudios/OpenSubdiv.git

Please advise.

Thanks.

jcowles:
Maybe I can fix this up for you, let me take a crack at it.

jcowles:
Merged.

Sheng, if you're curious, take a look at the commit history of OpenSubdiv/dev:
https://github.com/PixarAnimationStudios/OpenSubdiv/commits/dev

Notice it's only three commits. I did this by running 'git rebase -i upstream/dev'

Thanks again!

PR #700, opened 2015-07-13
Fixed initialization of unconnected vertices in FVar channels

barfowl:
Vertices in FVar channels were presumed to have at least one incident face.  This lead to some array bounds errors that were recently corrected (#695) but left the vertex uninitialized within the FVar channel.  This change corrects the initialization.

A regression shape with unconnected (bare) vertices was also added for future use.  Like other non-manifold cases, it cannot be included in the far_regression as Hbr will reject it.

Note that unconnected vertices will be implicitly filtered from any feature-adaptive refinement, so these cases are currently only relevant to uniform subdivision.


jcowles:
Hooray regression test!

PR #702, opened 2015-07-16
Fixes cmake projects for non-gpu use case

dazcjones:
Changes to the cmake projects so that if all the gpu options are disabled it does not try to make the gpu library.  Also made it so GLFW is not required if NO_OPENGL is specified.

Not sure if this was the best way to do this, but it seemed that the gpu build should not be required if the NO_* options don't require a gpu library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,8 @@ option(NO_DX "Disable DirectX support")
 option(NO_TESTS "Disable all tests")
 option(NO_GLTESTS "Disable GL tests")
 
+set(OSD_GPU FALSE)
+
 # Check for dependencies
 if(NOT NO_OMP)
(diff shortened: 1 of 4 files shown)


jcowles:
Thanks!
This makes sense, but it also looks like OpenGL ES is not handling the NO_OPENGL flag correctly.

That is, OpenGL is handled this way:
  if( OPENGL_FOUND AND NOT NO_OPENGL)
      set(OSD_GPU TRUE)
 
but OpenGL ES is as follows:
  if(OPENGLES_FOUND)
      set(OSD_GPU TRUE)

Perhaps still worth merging in it's current state, I'll see if there are any objections.

davidgyu:
This looks good to me, too. I'll verify the ES path after merging.

PR #703, opened 2015-07-19
Fixing linking errors on both Linux and WIndows

Nazg-Gul:
This pull request contains fixes which we needed to apply on the sources in order to make OpenSubdiv linking finely with Blender. Using gcc 4.9 on Linux and MSVC 2013 on Windows.

This is so called "works for us" and extra checks on whether it causes regressions for someone wouldn't hurt.


jcowles:
Thanks for the fix!

PR #704, opened 2015-07-20
Add ISPC limit surface evaluation

shengfuintel:
This pull request is to add ISPC based limit surface evaluation. The patch evaluation kernel is implemented by ISPC, an open source Intel® SPMD Program Compiler, and parallelized with TBB. To take the advantage of ISPC evaluation, the patch coordinates belongs to the same patch must be bundled together in path coordinates buffer. I have modified the demo glEvalLimit to bundle the patch coordinates. 

Ideally if we can modify struct PatchCoord from

PatchCoord{
    Far::PatchTable::PatchHandle handle; ///< patch handle
    float s, t;              ///< parametric location on patch
};

to 

PatchCoords{
    Far::PatchTable::PatchHandle handle; ///< patch handle
    int nCoord, 
    float *s, *t;              ///< parametric location on patch
};

We should save memory for patch coordinates buffer, in ISPC kernel, we should get better patch coordinates bundle since now the patch coordinates belongs to the same patch may be split into different threads, and I need to search the buffer to find the bundle. However, if we make this change, it will affect other kernels, which I am not familiar with. So I try not to change the data layout for now. 

The attached image is the performance data comparing ISPC kernel and TBB kernel. For single thread, the speed up is around 4 ~ 6 x faster. For multl-threaded. the speed up is about 1.3 ~ 2.1x faster. We get more speed up when we have more sampling points.


shengfuintel:
This is the performance data I collected:

ispc_performance

shengfuintel:
The following is how I run cmake to use ISPC evaluation:

/usr/pic1/cmake-2.8.9-Linux-i386/bin/cmake \
  -DCMAKE_C_COMPILER=/opt/intel/composerxe/bin/icc     \
  -DCMAKE_CXX_COMPILER=/opt/intel/composerxe/bin/icpc  \
  -DTBB_LOCATION=/opt/intel/composerxe/tbb             \
  -DCMAKE_BUILD_TYPE=RELWITHDEBINFO                    \
  -DGLEW_LOCATION=/usr/pic1/glew-1.9.0                 \
  -DGLFW_LOCATION=/usr/pic1/glfw-3.0.2                 \
  -DISPC_DIR=/rel/third_party/intelispc/1.8.1          \
  -DNO_PYTHON=1                                        \
  ..
  
You can download ISPC from https://ispc.github.io/downloads.html

takahito-tejima:
wow, this is a great addition! I see your point about PatchCoord struct. Bundling by patch makes a lot of sense if the client is ok to reorder the patch coords. Also we may want to have separate structs for CPU and GPU kernels, ideally... but I'm just not sure what kind of other use-cases we want to cover.
Your change is awesome, and we continue to discuss better interface for PatchCoord.
Thank you very much!

jcowles:
It looks like this change turns on C++11 for ICC, is that 100% necessary? 

If we've decided to flip that switch, we should do it for all compilers and ensure we're not breaking any clients who can't yet adopt C++11.

takahito-tejima:
My bad, Sheng, at this point OpenSubdiv is built without c++11, so we can't merge that change into dev. I should be more careful in merge.
Could you revise the evaluator to be compiled without c++11?

barfowl:
Glad to see the work on a separate evaluator for ISPC and looking forward
to eventually getting some kind of bundled evaluation in place.  While
we're looking into the C++11 issues, there was a minor public interface
issue that caught my eye...

I noticed the addition of a public isEqual() method to
Far::PatchTable::PatchHandle and had a couple of issues to raise about it:

1)  PatchHandle is a primitive struct with no methods and I wonder if any
should be added.  Usually the PatchTable invokes methods with the
PatchHandle as a parameter.  I didn't see any precedent for comparison
methods on such primitive structs, so I was wondering if operator==() was a
possibility, which would make it more STL friendly (and patchHandle is
likely to end up in STL containers at some point).

2)  If not, if it is to be a public method, it should be capitalized and so
named IsEqual().

3)  Given that it's likely called frequently, its implementation can also
be improved.  The third test of "vertIndex" should be redundant -- if the
Handle has the same patchIndex into the same array, then its guaranteed to
have the same set of vertices associated with it (the vertIndex doesn't
identify the Handle, its just a convenient and often retrieved datum
associated with it).  Making the patchIndex test first will then
short-circuit the test quicker for most cases -- most pairs of tested
patches will be in the same array, but few will have the same index.

That's a lot to say on a pretty minor issue. :)  There's so much more good
stuff here and I hope to dig into it deeper soon.

shengfuintel:
Turn on c++11 make it much easier to write TBB parallel for, I can use delta function instead of writing a new class. I only turned on when compiler is ICC. ICC definitely support c++11. Could it be an issue?

shengfuintel:
Thank Barry for your comment, I will fix it.

shengfuintel:
I believe the version of GCC required by OpenSubdiv also support C++11. Are there any other compilers we need to support?

jcowles:
It's both about compiler support and compatibility -- there are some ABI breaking changes in C++11 that can cause trouble for clients, so we want to be careful in terms of adoption and careful not to expose C++11 features in headers.

Our next release is focused on stability, so it seems appropriate to hold this change for a bit -- takahito has reverted it temporarily, but we all agree this is an awesome contribution and it's implemented the right way, so it's just a matter of timing.

Also, perhaps in the 3.1 release we could enable C++11 after careful consideration.

So please do address Barry's notes and if you would like to decouple this change from C++11 adoption, you could convert the lambdas, etc. 

Would you mind opening a new pull request? Then we can merge it at the appropriate time.

This is really great, thanks!!

Issue #694, opened 2015-06-30
FAR Tutorial 7 (Cascading Stencils) does not generated the correct result


dazcjones:
FAR Tutorial 7, that demonstrates the use of cascading stencils, does not seem to work correctly in release 3.0.  It applies the edit, but the resulting mesh seems to be scaled down or loses volume on each iteration of subdivision.

dazcjones:
Looking at it more closely, it looks like the StencilTableFactory::Create method ignores the "factorizeIntermediateLevels" options and always factorizes the weights of all levels.

jcowles:
Yes indeed, this is a bug -- thanks for reporting and digging into it!

We will have a fix soon.

--
Jeremy

dazcjones:
From my examination, the "factorizeIntermediateLevels" needs to be passed on to the internal "WeightTable" class from "stencilBuilder.cpp" or that "WeightTable" needs its "_coarseVertCount" member updated after each iteration.  Not sure which is the preferable solution or if there may be a better one, but those looked like the simplest and most localized.

dazcjones:
I see you have a "dev-stencilfix" in jcowles/OpenSubdiv, is that branch in a state that's worth pulling and trying?

dazcjones:
I tried the fix in "jcowles/OpenSubdiv/dev-stencilfix" it does seem to fix the issue.

jcowles:
Thanks for the verification, I'll pull request this soon.

--
Jeremy

jcowles:
I went with a simpler fix (updating coarse verts, one of the options you initially suggested) -- if you have time, would you mind testing again with this alternate fix (in PR #709)?

dazcjones:
The alternate fix seems to to work as well.

davidgyu:
This was fixed by PR#709

PR #709, opened 2015-07-23
Respect factorizeIntermediatLevels option

jcowles:
Previously, this option flag was being ignored during stencil table
construction. This change now respects the flag.

Fixes issue #694


jcowles:
I went for the simpler option here -- the more complex solution (in my dev-stencilfix branch) was attempting to encapsulate the details of stencil construction, but ultimately I didn't think it was worth added complexity.

PR #714, opened 2015-07-30
Replaced topology initialization assertions with errors

barfowl:
The Far::TopologyRefinerFactory requires a non-zero number of some components on initialization, i.e. the number of vertices and faces.  This had been detected with assertions, which have now been replaced with errors (actually warnings to be consistent with other reporting) allowing the TopologyRefiner creation to simply fail.


barfowl:
After a second look at this, I'm really questioning why Warning() was being used for similar errors instead of Error().  It may have had some history with the evolution of error codes used by Error(), but that doesn't seem relevant any more.  Given the conditions tested cause failure of the method, they really warrant the use of Error().

It's also been pointed out that the changes here had some redundant use of snprintf().  I'll deal with that and the above in an upcoming commit.

PR #719, opened 2015-07-31
Minor content and formatting changes to Far error strings

barfowl:
This change includes updates to Error() calls in Far so that their message strings are consistent.  All strings now include a specific (public) method within which the error originated along with its cause.

Minor changes to the formatting of the default message handlers were also made based on offline discussions.

### DIFF
--- a/opensubdiv/far/error.cpp
+++ b/opensubdiv/far/error.cpp
@@ -33,37 +33,43 @@ namespace OPENSUBDIV_VERSION {
 
 namespace Far {
 
+//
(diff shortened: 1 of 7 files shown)


barfowl:
Agreed.  There was debate as to what the future of the different types would be and there was only one instance of a CODING error in all libraries combined, so rather than introduce more I just switched it to RUNTIME.  It may be worth a comment in far/error.h that only one of the ErrorTypes is really used -- and may ever be used.

PR #723, opened 2015-08-03
Updated release notes for 3.0.1

davidgyu:


barfowl:
Is there a convention in these notes to distinguish between "Bug fixes" and
other pro-active improvements?  Many of the items under "Bug fixes" are the
latter.

davidgyu:
Good point!

I suppose a general rule is that it is a bug if it is likely to have caused
a support request, whether or not the code itself was incorrect. But does
the following seem like a more informative split? If so I can pull request:


https://github.com/davidgyu/OpenSubdiv/commit/20982d68b1a2bdd32c426b2eb403a6258672c867

Thanks!
-David

barfowl:
Agreed -- fixing a bug before its reported as a bug is still a bug fix.  I
went back to older release notes and see the that three categories of items
used were "New Features", "Changes" and "Bug Fixes" -- so the "Changes" in
your proposal makes sense.  Adding explicit casts to reduce compiler
warnings is still more a "Change" to me than a "Bug Fix" -- particularly as
the compiled code doesn't differ at all -- but I don't want to get pedantic.

I don't know how the rest of you feel about this, but I prefer to list bug
fixes in order of some level of significance.  Granted that's pretty
subjective and debatable, but the intent is to avoid having a really
significant fix buried within (and potentially obscured by) a bunch of
insignificant ones.  Something to consider if a resubmission of this is
coming...

Thanks

davidgyu:
Sounds good. I made a further change to move the remaining compiler warning
cleanup to the "Changes" section.

I didn't see a big advantage to re-ordering the bug-fixes, and the list for
this point release is relatively short anyway.

Otherwise, I think we're good to go for the release.

Thanks!
-David

PR #728, opened 2015-08-07
use WIN32_LEAN_AND_MEAN properly

significant-bit:
Misspelled W32_LEAN_AND_MEAN in osd/opengl.h had no effect on windows.h.

Also added to example code wherever windows.h is included.


jcowles:
Doh, thanks for the fix! 

Were you able to test the CL, CUDA and GLES builds to ensure everything is still working?

significant-bit:
Tested OpenCL and CUDA yes, didn't try GLES.

PR #734, opened 2015-08-24
Fixed small bug in glFVarViewer for Loop.

c64kernal:


davidgyu:
Looks great! Thanks George!

PR #738, opened 2015-09-10
Fix GLSL program being deleted twice in GLSL Compute controller

Nazg-Gul:
GLSL programs used by stencil and patch kernels were freed from
both GLSL Compute Evaluator and from kernel classes themselves.


jcowles:
Thanks Sergey! We'll review this and get it into the 3.0.3 stability release.

jcowles:
Also, just to confirm, does this PR fix issue 735?

takahito-tejima:
Ah! Yes! thank you for the fix.

PR #740, opened 2015-09-17
New "tutorial_8" calculates approximated smooth normals for uniform subdivision

poljere:
Added "tutorial_8" that shows one way of calculating approximated smooth normals for uniform subdivision.


barfowl:
Thanks for this.  I realize that this is example code, and only one way to compute normals, but if it is going to be perceived as "recommended"  I had a couple of minor questions/issues...

I noticed that for quads (the common case) only three of the four vertices are considered in the computed face normal that is accumulated with each of its vertices.  There are some cases where this bias to one side of a highly non-planar quad will be noticeable.  Using the cross product of the two diagonals of the quad will account for all four vertices.

If we are going to use the TopologyRefiner to help compute these, and we have full topology in the last level, then I wonder if there was a reason to avoid considering each vertex and its incident faces.  This approach is much more threadable than the current approach -- where the accumulation for each vertex inhibits threading.  It will compute more cross-products but will only normalize once.  It is a little more complicated topologically, so simplicity is a factor to consider.

poljere:
Yes, the idea with Tutorial_8 was to show ways to calculate approximate smooth normals. 

Currently, the code is very similar to what OSD 2.x was doing to calculate smooth normals, it is just an approximation using the first 3 verts of the face and accumulating them.

However, right after submitting this pull request we have realized we could also include code (in this same example) to show how to use the normals at the limit, so I am adding this as we speak. I think it will be useful for everyone to be able to play around with the different methods/results.

Regarding using the cross product of the two diagonals of the quad, it is a good idea, maybe we should also add it to the sample as another method to approximate smooth normals. Thoughts?

barfowl:
Personally I'm not a huge fan of associating limit normals with points that
aren't also on the limit, but I may be in the minority here.  A tutorial
that computes both limit points and normals is certainly worthwhile.  How
about factoring this example into a choice of two use cases:  refined
points and approximated normals, vs limit points and limit normals.

Regarding the quad issue, here's a pathological case:  vertex V has 4
incident faces, but is the last vertex in each face -- so the position of V
is never considered in the normal for V.  Given the way face vertices are
generated in refined levels, this may not be possible at anything other
than level 0, but lesser cases like this are still possible particularly
along boundaries.

jcowles:
Perhaps having three choices for normals will accentuate the fact that there is a choice to be made, further justifying its exclusion from the core?

A 2.x behavior example was explicitly requested, so I think it should be included, even if inferior.

As an extension of your objection, crossing the streams of limit and approximations is potentially confusing, but also potentially enlightening and convenient to see the different options implemented side by side?

I think the test shape should include some creases to highlight the errors.

--
Jeremy

poljere:
I have just added two of the methods we have talked about to the smooth normals tutorial. The first one is the one the Barry recommended, and the second one is using the normals from the limit.

barfowl:
Most of the other far tutorials seem to be using uniform refinement.  Why
don't we just add the normal approximations to tutorial_0 and make
tutorial_8 one that focuses on computing limit positions and normals?

There is a huge additional cost to limit positions and normals that should
be made clear.  For one thing, it requires the "full topology in last
level" option to be set, which adds a huge cost to the uniform refinement
(time and memory).  There's already a lot of confusion in the community
aside from our implementation, and combining too many concepts in one
example may not help that situation.

jcowles:
I think it's nice that tutorial_0 is super simple and it's also convenient that all the normal calculations are bundled together (if we one day gave the tutorials sane names, perhaps this one might be called "normals", etc).

However, I think you make a good point about confusion, could we just put a big banner comment in this tutorial explaining the costs and trade offs of the different techniques as well as noting the fact that limit normals should probably be used with limit positions?

barfowl:
The tutorial_0 may be so simple it has no practical use, but with a minor
addition (one small block of code) it becomes so.  Generating the faces for
a subdivided level with vertex positions AND normals is pretty straight
forward.  The message here is that a subdivided level is just another
polygonal cage, and here are some ways to put normals on a polygonal cage
(the 2.x and more quad-friendly versions).  The resulting .obj output is
now far more useful (viewable) and the example includes everything
necessary to make this now practical use case efficient.

Limit properties are another topic entirely, and the pros and cons of using
them can be detailed there.  Confusing that example with "smooth normals"
isn't helpful.  Once someone is familiar with both, they can decide how to
mix and match if they want (and limit positions with smooth normals has its
uses when the tessellation is relatively coarse).

In general I'm not in favor of a completely separate "smooth normals"
example.  As far as subdivision surfaces are concerned the only kind of
normal is a normal on the limit surface.  If users want to put them on
cages there are many ways to do so and we can show a couple in passing, but
a dedicated example makes it seem like we're making recommendations on the
topic.

I'll leave it and that and defer to others for now.  We've all agreed that
the tutorials warrant revisiting in a future release, so we can pick this
topic up there again if necessary.  Thanks

poljere:
I have just done another pass to the tutorial, this time I have added a big description on the different methods and the problems they may introduce, let me know what you think.

c64kernal:
Barry, you make excellent points here.  My reservations is that I think a tutorial_0 would be as simple as possible.  I worry if we add any kind of fake normals computation there, it will be seen as "how you do it" because after all, it's in the very first tutorial.  That makes me inclined to want to keep the normals computations isolated to one tutorial, where one could see the various ways of doing it and the implications of each.

And, underlying all of that, I feel, as I think we all do, we really need to revisit the tutorial structure and hopefully we can find a place where these snippets are easy to find and their purpose clear.

Let us know if you have any objections to landing as-is.

Thanks Barry!

jcowles:
I'm going to merge this in so we at least have something to share, but we can refactor as desired when we take a pass at reworking the tutorials.

PR #743, opened 2015-09-21
Fixed GLSL Compute Kernel buffer layout

davidgyu:
For improved portability, we need to declare std430 buffer layout
for buffer resources used by the GLSL Compute Kernel.


jcowles:
Since Anshu filed a PR, I'm waiting to hear back about the CLA before merging.

PR #744, opened 2015-09-21
Removing debug code that mistakenly slipped into glPatchTable.cpp.

takahito-tejima:
Also use GL_ARRAY_BUFFER for the binding point of updating indices instead of GL_ELEMENT_ARRAY_BUFFER


jcowles:
Thanks!

PR #745, opened 2015-09-23
Improved patch approximations around non-manifold features

barfowl:
Patches around non-manifold vertices are currently approximated with regular patches, which works well in some cases but not all.  Ultimately we need irregular patches (or at least some kind of extrapolated regular patch) to represent these without cracks, which is planned for the future.

Until then, this change uses regular corner patches around non-manifold vertices that are sharpened to interpolate the vertex and reduce cracks between the patches around it.  This does not eliminate cracks with the adjacent interior patches, but it is hoped the isolation level is deep enough to make them negligible until the longer term solution is in place.

### DIFF
--- a/opensubdiv/far/patchTableFactory.cpp
+++ b/opensubdiv/far/patchTableFactory.cpp
@@ -898,7 +898,10 @@ PatchTableFactory::identifyAdaptivePatches(AdaptiveContext & context) {
             }
 
             //  Identify boundaries for both regular and xordinary patches -- non-manifold
-            //  edges and vertices are interpreted as boundaries for regular patches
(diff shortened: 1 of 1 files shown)


jcowles:
It seems we could add better comments and add some symbol names for the magic numbers in PatchTable factory, but this change is just following the existing pattern.

It would be nice to clean this up in the future though, to lower the barrier to entry for new engineers...

barfowl:
Agreed.  When we deal with infinitely sharp features better (hopefully
soon), both the boundary and non-manifold cases handled here will be dealt
with by the general case of infinitely sharp -- that will simplify things
considerably.

PR #759, opened 2015-11-16
Teach the LimitStencilTable to compute 2nd partial derivatives

dcclyde:
When doing cloth simulations the 2nd partials are needed to describe the curvature.


dcclyde:
Oops, this should target the "dev' branch. One sec...

PR #760, opened 2015-11-16
Teach the LimitStencilTable to compute 2nd partial derivatives

dcclyde:
When doing cloth simulations the 2nd partials are needed to describe the curvature.

### DIFF
--- a/opensubdiv/far/patchBasis.cpp
+++ b/opensubdiv/far/patchBasis.cpp
@@ -208,7 +208,7 @@ inline void Spline<BASIS_BILINEAR>::GetPatchWeights(PatchParam const & param,
         derivT[2] =   s * dScale;
         derivT[3] =  sC * dScale;
 
-        if (derivSS and derivST and derivTT) {
+        if (derivSS && derivST && derivTT) {
             float d2Scale = dScale * dScale;
 
             for(int i=0;i<4;i++) {
@@ -291,7 +291,7 @@ void Spline<BASIS>::GetPatchWeights(PatchParam const & param,
             }
         }
 
-        if (derivSS and derivST and derivTT) {
+        if (derivSS && derivST && derivTT) {
             // Compute the tensor product weight of appropriate differentiated
             // (s,t) basis functions for each control vertex (scaled accordingly):
         
--- a/opensubdiv/far/patchBasis.cpp
+++ b/opensubdiv/far/patchBasis.cpp
@@ -46,14 +46,14 @@ class Spline {
 public:
 
     // curve weights
-    static void GetWeights(float t, float point[], float deriv[]);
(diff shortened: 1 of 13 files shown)


dcclyde:
These changes are from WDAS so the CLA is already handled.

barfowl:
Wow, this is great timing, thanks.  We had other needs for 2nd derivatives
and I was slated to start working on the addition before the end of the
year, so this will save me some work.

We did have some open questions on how to extend the API to support 2nd
derivatives when the topic came up in the past, so that may become a
discussion point here.  I haven't had a close look yet but am really glad
to see this addition on its way.

Thanks again!

barfowl:
I had a look at some of the proposed changes before the holidays, and something in particular caught my eye that may warrant a separate discussion...

In far/patchBasis.cpp you swtiched the #ifdef controlled by _USE_BEZIER_PSEUDO_DERIVATIVES so that the true derivatives of a Gregory patch are used, along with an accompanying comment that included "skipping half of the product rule like this does seem to change the result a lot in my tests".

Can you identify or extract any specific cases that would clearly illustrate the benefits of using the true Gregory derivatives?  I used contrived Gregory patches to test the 1st derivative evaluation and the differences with cases arising around typical extra-ordinary vertices were subtle, but I suspected that any difference would be compounded with higher order derivatives and may eventually cause trouble (and was planning to investigate when the time came for those higher derivatives).

I actually favor using the true Gregory derivatives instead of the Bezier approximations (with a little more factoring of the evaluation the added computation expense should be minor), but all of the shader code for Gregory patches also uses the approximations.  So I left the approximations in place here hoping we would change them all at some point in the future.  Changing just the code in patchBasis.cpp as done here will mean we have inconsistent 1st partials and normals between CPU and GPU implementations.

So I'd like to make the case for us switching to accurate evaluation of Gregory patch derivatives across the board, and that case will be better made with some examples that illustrate the benefits.

Thanks

davidgyu:
Hi Dave,

We're at a point where we would love to merge this PR. There are a couple of merge conflicts resulting from our earlier changes to eliminate the use of alternative keywords. A few files need 'and' changed to '&&'. Let me know if you'd like to update the source branch, otherwise I can resolve the conflicts.

Thanks!
-David

dcclyde:
Is this the update you intended? I'm happy to do something differently if I got it wrong. I use git a lot but I don't have much experience with github.

davidgyu:
Thanks for the update! Yes, I think this looks good. -David

DarkDefender:
@davidgyu will this functionality become available via the `EvalPatches` function?

davidgyu:
Yes, we are planning to support this across the evaluation interfaces in OpenSubdiv.

PR #772, opened 2016-01-05
Add events control to clVertexBuffer and clEvaluator

xiao-deng:
Allow giving the start events and returning the end event when updating and evaluating the vertex buffer. So it will be more flexible to do synchronization for parallel execution of multiple CL kernels.

If no event is given, the methods will behave the same as before.


jcowles:
CLA received, thanks! Someone will review this soon.

takahito-tejima:
the change looks great! thank you.

davidgyu:
Looks good to me too!

PR #781, opened 2016-02-01
Tess shader performance fix

davidgyu:
These changes help improve the measured performance of the
GL tessellation control shader code and mitigate some performance
regressions we've seen since v3_0_0_RC1

Equivalent changes have been made to both the GLSL and HLSL
code paths. Performance is improved somewhat on DX11 but the
differences there are much less than GL.


takahito-tejima:
We reviewed this change in person. It looks good and I'm glad to see the performance is recovered.
thanks!

Issue #776, opened 2016-01-25
Generate MSCV solution with static CRT?


thomthom:
Is it possible to have CMake generate a MSVC solution that uses static CRT? (/MT)

I'm admittedly not that familiar with CMake and I'm not sure if this is something that is generic to using CMake or if the project has to build in support for it.

My use case is that I'm writing an extension for an application where I'm implementing OpenSubdiv. The extensions for this application normally doesn't use installers but instead are packaged as simple ZIP packages that is unpacked. Hence I'm trying to avoid the need to depend on CRT.

jcowles:
A cursory search seems to suggest it's something you must configure manually, but I haven't checked our cmake setup to see if it was implemented previously.

--
Jeremy

thomthom:
Configured in Visual Studio after CMake generation, or can I configure CMake to generate the solution up front?

c64kernal:
Hi Thomas, I'd recommend you configure this in cmake and not modify the generated projects:

https://cmake.org/Wiki/CMake_FAQ#How_can_I_build_my_MSVC_application_with_a_static_runtime.3F

jcowles:
+1 and if you get it working via CMake, please send a pull request!

--
Jeremy

jcowles:
Oh, it looks like the doc George sent is explains how to setup a one-off CMake override, which is probably not something we would distribute.

But if you get it working for real with a config option, then a pull request would be great.

--
Jeremy

thomthom:
Will do!

manuelk:
I vaguely remember early releases of OSD having the /Mt option, but it looks like it's disappeared.

This *should* be as easy as inserting /Mt here:
https://github.com/PixarAnimationStudios/OpenSubdiv/blob/master/CMakeLists.txt#L272

Although : why not link statically both OSD and CRT with your shipping binary ?

thomthom:
> Although : why not link statically both OSD and CRT with your shipping binary ?

That's what I'm trying to do - but then I need to build the OSD libs with `/Mt` to match.

Thanks for the hint at where to change the `CMakeList.txt` - I'll be digging into it as soon as I get a spare moment.

thomthom:
> This should be as easy as inserting /Mt here:
That would change it permanently to be static CRT, right? Wouldn't it be desired to have this be configurable when running cmake?

thomthom:
[...]

PR #783, opened 2016-02-10
Static CRT for MSVC (DON'T MERGE THIS)

thomthom:
I don't intend this to be merged - but merely as a way to talk about the changes I had to do to solve issue #776. Got some questions - once resolved I'll create a new clean pull request.

Testing with this config:

```
cmake ^
  -G "Visual Studio 12 2013 Win64" ^
  -D NO_EXAMPLES=1 ^
  -D NO_TUTORIALS=1 ^
  -D NO_REGRESSION=1 ^
  -D NO_MAYA=1 ^
  -D NO_PTEX=1 ^
  -D NO_DOC=1 ^
  -D NO_OMP=1 ^
  -D NO_TBB=1 ^
  -D NO_CUDA=1 ^
  -D NO_OPENCL=1 ^
  -D NO_OPENGL=1 ^
  -D NO_DX=1 ^
  -D NO_CLEW=1 ^
  -D MSVC_STATIC_CRT_RELEASE=1 ^
  -D MSVC_STATIC_CRT_DEBUG=1 ^
  ..
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,10 +273,10 @@ elseif(MSVC)
     )
 
     # Turn off a duplicate LIBCMT linker warning
-    set(CMAKE_EXE_LINKER_FLAGS
-        "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
-    set(CMAKE_SHARED_LINKER_FLAGS
-        "${CMAKE_SHARED_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
+    #set(CMAKE_EXE_LINKER_FLAGS
+    #    "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
+    #set(CMAKE_SHARED_LINKER_FLAGS
+    #    "${CMAKE_SHARED_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
 
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,19 @@ option(NO_DX "Disable DirectX support")
 option(NO_TESTS "Disable all tests")
 option(NO_GLTESTS "Disable GL tests")
 
+option(STATIC_CRT "Statically link MSVC CRT" OFF)
+
+if(MSVC AND STATIC_CRT)
+    message(STATUS "Using static MSVC CRT")
+    # http://stackoverflow.com/a/32128977/486990
+    add_compile_options(
+        "$<$<CONFIG:Debug>:/MTd>"
+        "$<$<CONFIG:RelWithDebInfo>:/MT>"
+        "$<$<CONFIG:Release>:/MT>"
+        "$<$<CONFIG:MinSizeRel>:/MT>"
+    )
+endif()
+
 set(OSD_GPU FALSE)
 
 # Check for dependencies
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,10 @@ elseif(MSVC)
                     #/D_HAS_ITERATOR_DEBUGGING=0
     )
 
+    # thomthom : When enabling static CRT there are linker errors in Release
+    # referring to libcpmt (Note not libcmt). Not sure why, but removing this
+    # exclude allowed it to build, and there where no linker error in VS2013/15
+    # when building with minimal build.
     # Turn off a duplicate LIBCMT linker warning
     #set(CMAKE_EXE_LINKER_FLAGS
     #    "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -272,11 +272,15 @@ elseif(MSVC)
...
+    # thomthom : When enabling static CRT there are linker errors in Release
+    # referring to libcpmt (Note not libcmt). Not sure why, but removing this
+    # exclude allowed it to build, and there where no linker error in VS2013/15
+    # when building with minimal build.
     # Turn off a duplicate LIBCMT linker warning
-    set(CMAKE_EXE_LINKER_FLAGS
-        "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
-    set(CMAKE_SHARED_LINKER_FLAGS
-        "${CMAKE_SHARED_LINKER_FLAGS} /NODEFAULTLIB:libcmt.lib")
+    #set(CMAKE_EXE_LINKER_FLAGS
```
thomthom: This is my biggest concern. I don't understand why I was getting `libcpmt.lib` linker errors only in Release. And why the `cp` version and not the `c` version which is actually excluded.

Linker error log: https://gist.github.com/thomthom/5f9a23263c6a05ec5c5d

I did not observe any linker warnings by removing this. But maybe that's because I tried with a minimal config.

Anyone got any ideas to what is going on here?

CMakeLists.txt
```diff
@@ -314,6 +318,25 @@ option(NO_DX "Disable DirectX support")
 option(NO_TESTS "Disable all tests")
 option(NO_GLTESTS "Disable GL tests")
 
+option(MSVC_STATIC_CRT_RELEASE "Statically link MSVC CRT in Release" OFF)
```
thomthom: Not sure what option name you'd prefer. This naming convention goes against the existing `NO_` options. Could invert them if preferable. `NO_DYNAMIC_MSVC_CRT` etc.

CMakeLists.txt
```diff
@@ -314,6 +318,25 @@ option(NO_DX "Disable DirectX support")
...
+if(MSVC AND MSVC_STATIC_CRT_RELEASE)
+    message(STATUS "Using static MSVC CRT Release")
+    # http://stackoverflow.com/a/32128977/486990
+    add_compile_options(
+        "$<$<CONFIG:RelWithDebInfo>:/MT>"
+        "$<$<CONFIG:Release>:/MT>"
+        "$<$<CONFIG:MinSizeRel>:/MT>"
+    )
+endif()
+if(MSVC AND MSVC_STATIC_CRT_DEBUG)
```
thomthom: Also not sure if it's desirable to have separate config for Release and Debug on this. The reason I did this was that in the current project I have I actually have dynamic CRT in Debug and static CRT in Release because I needed static only for released products. I have no strong attachments to this setup and I can simplify to make the flag affect both configs.

CMakeLists.txt
meshula: It would be weird to invert them IMO.

CMakeLists.txt
meshula: Ideally you could set things up so all four possible variants are available in the vcxproj, rather than having to re-run cmake again and again to get them all.

thomthom: I'm not sure if I follow you? _Four_ variants?
You only pick between static and dynamic though. And I was thinking you pick the setup for your build - once. Not sure why you want to build every pre-mutation?

meshula: I see your point, if you're building a single variant into your project, and the build files for OSD are local to your project that makes total sense.

The four variants are Static-Debug, Static-Release, Dynamic-Debug, Dynamic-Release. I don't think we have to worry about MultiThreaded or non-MT any more (so thankfully there aren't 8 variations).

In a larger build ecology it's common to build all the variants for packaging so that individual programmers don't have to do the builds at their desks. Of course one could re-run cmake to generate all the variants and do builds but that's far less convenient than having all the variants in a single project file because one only has to run msbuild four times.

thomthom: Gotcha. Makes sense. All though, I'm not entirely sure how to set up CMake to do that.
That means adding additional configs so you'd have `Release`, `Release Static`, `Debug`, `Debug Static` etc?

There's another aspect of this as well, 32bit vs 64bit. At the moment I am working on an application plugin which has both 64bit and 32bit builds. I've yet to find a way to make CMake generate a single solution that include both versions - like you'd normally have when manually crafting.

My current way around this is to create two build directories `build32` and build64` and then trigger CMake and msbuild. If there is a way to generate all in one solution that'd be great.

Btw, having all these CRT configs in one file - can this be done in a separate issue? I agree that it's be really convenient, but at the moment I'm not familiar with how to do that - and not having static CRT is somewhat of a blocker for my project.

meshula:
LOL, good point! Nothing worse than the tiny tweak that expands into a re-engineering effort :)

thomthom:
giphy

I don't mind doing some research into figuring it out. Good opportunity for me to learn Cmake better. But being able to do that separately would be good.

My main concern about this current PR is the `libcmt.lib` issues I had. I never saw the `LIBCMT` linker warnings that the comment mentioned - so I've been unable to verify if commenting out those lines causes a regression. If anyone has some insight into that I'd appreciate it.

thomthom:
Seeing how there seem to be activity on this repo these days, may I try to bump this? Static CRT would be really handy for me - a must in fact. I do get around it for the moment by building of my custom mod - but I need to reinstate this every time I sync up.

davidgyu:
Yes, I'll take a look...

thomthom:
Bumping this again. I wanted to update OSD, but I really need static CRT. I can get around this for now my modifying my local fork, but it'd be nice if there was an option for this.

davidgyu:
Cherry-picked and updated as part of PR #917 

Thanks again for getting this started!

PR #794, opened 2016-03-16
fix ptex lookup shaders to clamp the patchcoord

takahito-tejima:
with MSAA enabled, patchcoord could overrun the guttering pixels.


davidgyu:
Looks good to me and passes our internal test cases.

PR #795, opened 2016-03-16
Fixed residual memory consumed by end cap stencils

davidgyu:
An earlier change improved transient memory usage during the construction
of end cap stencil tables but could leave unused capacity in the internal
containers of the resulting tables. This change adds an additional
internal factory helper method which trims this storage.


takahito-tejima:
The change looks great to me. Thank you for fixing the regression!

Issue #789, opened 2016-02-20
Crash in MSCV Debug model with corner crease and CREASE_CHAIKIN


thomthom:
Yes - this is very specific - bear with me.

The crash happen under these conditions:
* Visual Studio 2013 compiler. Crash happen only in Debug mode - Release work fine.
* You have a creased vertex at the border of a mesh. Ie, not a solid manifold. Crease a vertex of a single quad.
* You have `CREASE_CHAIKIN` creasing mode.
* The Crease weight is less than or equal to the max refinement level. Crease weight higher than the refinement level works fine.

![image](https://cloud.githubusercontent.com/assets/192418/13197819/decef432-d7f8-11e5-9165-0071acda0460.png)

The crash occurs when `ComputeVertexVertexMask` returns. If you step through the code and use StepInto at the end it'll step into the destructor of `LocalMask` and `Crease` before it with crash inside the CRT (I think).

I created a gist with modified `far_tutorial0.cpp` which reproduce it as simple as I can manage: https://gist.github.com/thomthom/242aa26f8f469a05d5a5

I built OpenSubdiv using this Cmake config:

```
cmake ^
  -G "Visual Studio 12 2013 Win64" ^
  -D NO_REGRESSION=1 ^
  -D NO_MAYA=1 ^
  -D NO_PTEX=1 ^
  -D NO_DOC=1 ^
  -D NO_OMP=1 ^
  -D NO_TBB=1 ^
  -D NO_CUDA=1 ^
  -D NO_OPENCL=1 ^
  -D NO_OPENGL=1 ^
  -D NO_DX=1 ^
  -D NO_CLEW=1 ^
  ..
```

I don't know if this is an OSD bug or a MSVC Debug CRT bug - but I've reached as far as I can on this. I don't see an obvious cause or fix. But it's making it hard for me to debug my own application when some of my models keep crashing.

I am using the `dev` branch which is currently at a state where it just merged the 3.0.4 release - no other commits after that.

barfowl:
I'll have a look at this shortly since I am most familiar with that code.  I don't have access to MSVC but I'll see if there's anything suspicious with the Chaikin implementation in general.  There may be some corruption of the stack earlier on that is triggering this crash on return.

thomthom:
That's what I was wondering - bit strange that I only see this in Debug - but maybe that's just by chance so far. Maybe I see this in Debug because of how it handles memory differently and got more error checking.

[...]

PR #798, opened 2016-03-31
Fixed Chaikin-related bug causing crashes in debug for MSVC and others

barfowl:
This pull request addresses issue #789.

The bug occurs using the Chaikin crease method when a vertex transitions from Corner to Crease with infinitely sharp edges.  When no semi-sharp edges are present around a vertex, all child sharpness values should be inherited from their parent -- both smooth and infinitely sharp -- and not all presumed smooth.

### DIFF
--- a/opensubdiv/sdc/crease.cpp
+++ b/opensubdiv/sdc/crease.cpp
@@ -168,7 +168,7 @@ Crease::SubdivideEdgeSharpnessesAroundVertex(int          edgeCount,
         //
         if (sharpCount == 0) {
             for (int i = 0; i < edgeCount; ++i) {
-                childSharpness[i] = Crease::SHARPNESS_SMOOTH;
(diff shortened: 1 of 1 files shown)


davidgyu:
This looks good Barry. It would be good to have a mesh in our regression shapes which exercises this code path. Thanks! -David

barfowl:
Agreed -- that's what hung me up getting the pull request out last month.  Thomas' example really helped to track down the problem quickly (thanks again) but seemed too simple for a regression shape, which tend to illustrate a particular feature or two.  I was wondering what kind of new shape might be useful.

Keep in mind that the chaikin shapes are also all disabled in the regression tests due to their differences with Hbr, so that limits the effectiveness of adding any new shape.  Probably the simplest thing to do is sharpen some boundary vertices on one of the existing shapes...

PR #799, opened 2016-04-05
Improvements to far_regression

barfowl:
The changes included here are intended to give far_regression more direct control of each Shape object so that it can make appropriate decision based on content, potentially modify the Shape before execution, etc.  Global functions in regression/common were overloaded to take a pre-existing Shape object instead of the shape string to be parsed.  Only additions were made to the interface of regression/common, no changes to any function signatures.

While the intent here was to facilitate future extensions, the immediate effect on far_regression is as follows -- it will now:

    - load all regression shapes, not a subset based on assumptions of their content
    - refine all regression shapes with Far
    - inspect features of each shape for compatibility with Hbr and skip any deemed incompatible
    - refine compatible shapes with Hbr and compare results for these between Far and Hbr

So shapes with Chaikin creasing or non-manifold topology will be loaded and refined now -- just not compared for equivalence with Hbr.

### DIFF
--- a/regression/common/far_utils.h
+++ b/regression/common/far_utils.h
@@ -137,19 +137,16 @@ InterpolateFVarData(OpenSubdiv::Far::TopologyRefiner & refiner,
 
 template <class T>
 OpenSubdiv::Far::TopologyRefiner *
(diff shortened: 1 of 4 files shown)


davidgyu:
These are nice improvements, Barry. Thanks -David

Issue #779, opened 2016-01-29
MSVC : 'and'/'not' : undeclared identifier (error C2065)


nyue:
I am able to build the OpenSubdiv 3.0.3 library (no graphics, just the CPU evaluator stuff) on Windows (VS 2012)

When I attempt to use it in a project, I get the 'and' and 'not' keyword undefined.

I tried looking at the CMakeLists.txt files in the github source but could not find any obvious way to tell MSVC to honor the C++ standard keywords.

Cheers

c64kernal:
Hi Nicholas,

In the CMakeLists.txt file at the root of the project is a bunch of defines that are meant to solve this problem.  This is super hacky and we should clean this up.  Alternative tokens are provided by the iso646.h header and we should be using that instead.  The /Za option which disables MSVC's language extensions and allow it to support alternative tokens natively is not at all practical.

thomthom:
I ran into the same issue when making a VS project using OpenSubdiv. I worked around it by including `iso646.h` before including the OSD headers.

However, I meant to open a ticket on this myself. While I did find the comments in the `CMakeLists.txt` file a couple of days ago describing the issue, I wanted to suggest that either OSD include `iso646.h` or that perhaps the `and`, `or` and `not` keywords where refactored to the more common operators. 

I was confused at first when I ran into this - because the VS solution Cmake generates builds fine, but my own failed when I used the includes. I didn't realize that was these special project properties defined to address this VS compiler limitation.

By the way - is this project using any particular code style guide?

thomthom:
I'm also wondering if there are other project compiler flags that might benefit from being defined in the source?

```
                    # Make sure WinDef.h does not define min and max macros which
                    # will conflict with std::min() and std::max().
                    /DNOMINMAX

                    # Make sure the constants in <math.h> get defined.
                    /D_USE_MATH_DEFINES
```

thomthom:
What is the desired fix for this issue? Ensure `iso646.h` is included in the files that use these keywords? Or refactor to use `&&` and `||` etc instead?

c64kernal:
Hi Thomas, I think the general preference is to use the iso646.h header -- however there was some concern that it may not be supported on older compilers that we'd still like to support.

As far as general coding style, the documentation simply says: inspired by the Google style guide (http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml)

thomthom:
[...]

PR #807, opened 2016-06-25
Dev iso646 fix dx Ref #779

thomthom:
Renaming `and`, `or`, and `not` et. al for the DirectX project.


thomthom:
Made the PR against the wrong branch.

PR #817, opened 2016-07-07
iso-646 fix remove obsolete compiler defines

### LINKED ISSUE #779: MSVC : 'and'/'not' : undeclared identifier (error C2065)

davidgyu:
Now that all of the code changes have been integrated
we can remove the alternative keyword defines from
the MSVC build.

Fixes #779

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,42 +207,6 @@ elseif(MSVC)
     # Turn on all warnings
     list(APPEND OSD_COMPILER_FLAGS /Wall)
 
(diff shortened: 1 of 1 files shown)


c64kernal:
Awesome!  Thanks David!

PR #818, opened 2016-07-09
No longer use CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR

### LINKED ISSUE #814: CMakeLists: use OpenSubdiv_SOURCE_DIR instead of CMAKE_SOURCE_DIR
Hello there,

I have a suggestion regarding the cmake scripts. I noticed that the CMAKE_SOURCE_DIR variable is used for establishing paths which can be problematic if you are incorporating OpenSubdiv build in your personal project using add_subdirectory(). Unless you have a reason for using CMAKE_SOURCE_DIR I would recommend using <Project_Name>_SOURCE_DIR instead. This would make OpenSubdiv more compatible with various dependancy management policies. For example, I like to subtree/submodule dependancies at a particular commit so that all collaborators are working with the same development environment. 

Here is a link to the documentation:
[CMAKE 2.8.6 <Project_Name>_SOURCE_DIR](https://cmake.org/cmake/help/v2.8.6/cmake.html#variable:Projectname_SOURCE_DIR)

Here is an example of it being used by GLFW:
[GLFW Example](https://github.com/glfw/glfw/blob/3.2/CMakeLists.txt)

c64kernal:
As mentioned in issue #814, avoiding the use of CMAKE_SOURCE_DIR makes it easier to integrate OpenSubdiv as a submodule in other projects.

Fixes #814.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,12 +31,12 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 #-------------------------------------------------------------------------------
 # Obtain OpenSubdiv API version from version.h file
-if(EXISTS "${CMAKE_SOURCE_DIR}/opensubdiv/version.h")
-    file(STRINGS "${CMAKE_SOURCE_DIR}/opensubdiv/version.h"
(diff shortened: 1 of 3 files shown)


davidgyu:
Looks great!

PR #820, opened 2016-07-14
Fixed bug in Far::PatchTableFactory to correct sharpness assignment:

barfowl:
The bug appears to be the result of a cut/paste oversight -- when populating the irregular patch types, the sharpness index vector that was used was that of the regular patches rather than the specific patch type.

This doesn't appear as a problem now because the corruption of the regular vector only occurs in the last level when Gregory patches are generated, and single crease patches are not generated in the last level.  Once Gregory patches are generated before the last level, the corruption of single crease patches becomes evident.


davidgyu:
Looks great! Thanks Barry.

PR #821, opened 2016-07-14
Improved Vtr::Level::isSingleCreasePatch() test

barfowl:
At the time the original single-crease patch test was written, the tags for vertices were still evolving.  The vertex tags (VTags) were specifically designed to be bitwise combined so that the combined result of the corners of a face could be used to quickly deduce properties relevant to patches.  In this case we can get almost everything we need from the four corner tags without any topology traversal.

I kept the new version functioning the same as the old, but the context in which this test is usually run has the combined corner tags available and will already avoid many of the cases rejected here.  So there may eventually be merit in splitting this test in two:  the first working directly with available tags, the second inspecting sharpness values for the potential sharp edge of the face.

### DIFF
--- a/opensubdiv/vtr/level.cpp
+++ b/opensubdiv/vtr/level.cpp
@@ -1277,89 +1277,64 @@ Level::gatherTriRegularCornerEdgePatchPoints(Index fIndex, Index points[], int c
 }
 
 bool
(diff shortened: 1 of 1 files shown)


davidgyu:
Looks good.

PR #823, opened 2016-07-19
Add internal support for separate levels of refinement for different features

barfowl:
This change includes internal support allowing different levels of adaptive refinement for extra-ordinary vertices vs other sharp features -- preventing the over-representation of smooth patches around all extra-ordinary vertices when a high level of refinement is required for a particular crease.

The feature is not publicly exposed here yet (we need to work out some public interface details) but removes internal assumptions when constructing patches that irregular patches can only exist at the last level of refinement.  Any further changes in this area should now be constrained to the refinement stage rather than patch construction.  Changes to adaptive refinement here are minor but due to change more extensively soon to better support infinitely sharp creases.


davidgyu:
This looks great, Barry! Thanks! -David

PR #825, opened 2016-07-20
Update BSpline end cap factory to detect partial rings at corners

barfowl:
With the recent addition of VSpans to indicate use of a partial ring around a vertex, the BSpline end cap factory code needed to be updated to defer to Gregory patches in more cases.  Boundary patches may now be created at interior vertices when the VSpan at a corner indicates a subset of the neighborhood is to be used.


davidgyu:
Looks good to me. Thanks! -David

PR #826, opened 2016-07-20
Added new vertex tags to identify infinitely sharp features

barfowl:
As a basis for upcoming work to improve patches around infinitely sharp features, this change includes the definition and propagation of new vertex tags that identify the presence and nature of infinitely sharp features at a vertex.  As with other VTags, this greatly simplifies the logic required in adaptive refinement and patch construction.  The new tags are not yet used to affect adaptive refinement or patches -- they are only defined, initialized and propagated through the refinement hierarchy, making them available for these processes in future.


davidgyu:
Also, looks good to me. Thanks! -David

PR #828, opened 2016-08-05
Refactored Far::PatchTableFactory

davidgyu:
Re-organized the implementation of the patch table factory
in advance of adding support for bicubic face-varying patches.

- Renamed the AdaptiveContext internal class to BuilderContext
  since this can be used (eventually) to build patch tables for
  uniformly refined topology as well as adaptively refined
  topology.
- Added utility methods to gather patch vertices from the
  levels (and eventually fvarLevels) of a refiner.
- Replaced the FVarChannelCursor with a simpler array of integer
  face-varying channel indices.
- Extracted the topology inspection needed to identify
  patches into a computePatchTag() method.
- Simplified the two-pass traversal of the topology into a
  single pass traversal of the refined topology followed by
  an iteration over the patches that have been identified.
- Replaced internal templated PatchTypes class with simpler
  and more direct accounting and patch array building facilities.
- Changed the API of the end cap patch factories to no longer
  depend on the patch table factory's PatchFaceTag. This has
  been superseded by the specification of appropriate
  VSpans at the corners of the end patch.

### DIFF
--- a/opensubdiv/far/endCapBSplineBasisPatchFactory.cpp
+++ b/opensubdiv/far/endCapBSplineBasisPatchFactory.cpp
@@ -78,7 +78,6 @@ ConstIndexArray
 EndCapBSplineBasisPatchFactory::GetPatchPoints(
     Vtr::internal::Level const * level, Index thisFace,
     Vtr::internal::Level::VSpan const cornerSpans[],
-    PatchTableFactory::PatchFaceTag const * /* levelPatchTags */,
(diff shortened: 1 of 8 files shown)


barfowl:
Looks good -- glad to see this taking shape.  I'll leave you to change the spelling of "PatchTuppleVector" in a later set of changes... ;)

davidgyu:
Ah, good catch! I'll fix that. Thanks!

PR #835, opened 2016-08-25
Refactored Far::PatchParam

davidgyu:
Improved documentation for Far::PatchParam and
refactored the implementation to allow better reuse
for evaluating patches from face-varying channels.

These changes preserve backward source compatibility
with the public APIs.


barfowl:
A little more documentation describing the motivation and distinction between PatchParam and PatchParamBase would be helpful.  Let's consider adding that later -- no reason to hold up the request for now.

PR #836, opened 2016-08-25
PatchTableFactory changes for smooth FVar patches

davidgyu:
These changes add basic support for building patch
tables with smooth face-varying patches. There is more
work to do to improve topological inspection and
tagging in order to take full advantage of improvements
for handling the change of basis for irregular patches.


barfowl:
We should make a point to revisit the BSpline end-cap factory so that it doesn't defer to Gregory for all irregular face-varying patches.

PR #841, opened 2016-09-04
Enabled point sharing for face-varying Gregory patches

barfowl:
Sharing of points for adjacent face-varying Gregory patches had been disabled as edges that were discontinuous in face-varying space were not being detected in the Gregory patch factory.  The Gregory patch factory now detects such edges and the sharing option has been re-enabled in the PatchTableFactory.


davidgyu:
Great! Thank you Barry!

PR #842, opened 2016-09-06
Bilinear endcaps

manuelk:
First pass at adding a stencil factory for bi-linear end-caps. It seems to behave reasonably well in my characteristic maps branch. With some of the Gregory basis refactoring done, it would be probably good to have it for reference for review & upcoming code cleanups.

Ex.:
- passing an int for level index instead of a Vtr::level to GatherPatchPoints
- renaming GregoryBasis to something more generic, since it is used un-related places
- generalize the Masks class (is that part of the Sdc::SchemeWorker ?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,9 @@ before_script:
    - sudo apt-get install cmake
    - cmake --version
 
-   # install GL related libs from utopic
-   - sudo add-apt-repository 'deb http://us.archive.ubuntu.com/ubuntu/ utopic main restricted universe multiverse'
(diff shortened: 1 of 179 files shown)


davidgyu:
Hi Manuel,

Can you re-open this against the dev branch?

Thanks!
-David

manuelk:
D'oh...

thomthom:
One can change the branch for a PR now: https://github.com/blog/2224-change-the-base-branch-of-a-pull-request

manuelk:
Thanks for the tip - hopefully i won't have forgotten the next time i make the same mistake...

PR #843, opened 2016-09-06
Adding bilinear end-cap stencil factory

manuelk:
First pass at adding a stencil factory for bi-linear end-caps. It seems to behave reasonably well in my characteristic maps branch. With some of the Gregory basis refactoring done, it would be probably good to have it for reference for review & upcoming code cleanups.

Ex.:
- passing an int for level index instead of a Vtr::level to GatherPatchPoints
- renaming GregoryBasis to something more generic, since it is used un-related places
- generalize the Masks class (is that part of the Sdc::SchemeWorker ?)

### DIFF
--- a/opensubdiv/far/CMakeLists.txt
+++ b/opensubdiv/far/CMakeLists.txt
@@ -26,6 +26,7 @@
 # source & headers
 set(SOURCE_FILES
     error.cpp
+    endCapBilinearBasisPatchFactory.cpp
     endCapBSplineBasisPatchFactory.cpp
     endCapGregoryBasisPatchFactory.cpp
     endCapLegacyGregoryPatchFactory.cpp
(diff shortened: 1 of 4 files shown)


jtran56:
Filed as internal issue #151986.

barfowl:
This has now been addressed with the changes merged in #966.  The end-cap factories and GregoryBasis code has all been replaced.  The CatmarkPatchBuilder now supports conversion of refined points to the different patch types, and the linear case is now supported.

While the choice of Linear end-caps does now work for vertex patches, additional work is still necessary in the PatchTable to support it for face-varying patches (combining BSpline and linear patches in the same array is problematic).  An API extension is also necessary in Osd::MeshBits to enable the choice at a higher level, so full support won't be available until then.

Thanks

PR #846, opened 2016-09-07
Removed mayaPolySmooth example

davidgyu:
Also removed the remaining Maya dependencies. These
were most useful during the early days of OpenSubdiv.
Now OpenSubdiv has been integrated directly into Maya
and these examples serve little purpose.


c64kernal:
Awesome.  Thank you sir!

PR #850, opened 2016-09-16
Minor updates to tags for face-varying and inf-sharp topology

barfowl:
This change set includes some cleanup and minor extensions to the tags that are used to accelerate topology inspection.  Some obsolete/redundant tags were removed along with some awkward methods formerly used for their inspection.  Recent additions now provide a more consistent interface for gathering the tags for both vertex and face-varying topology.


c64kernal:
Takahito and I reviewed this, thanks Barry!

PR #852, opened 2016-09-23
Extended adaptive refinement for inf-sharp patches and simplified

barfowl:
The motivation for this change was the addition of a new adaptive refinement feature to stop isolation of inf-sharp patches when they are detected as regular.  Given other recent changes to adaptive refinement (separate isolation levels), the need to support backward compatible behaviour, the addition of new tags to simplify selection, etc., it was time to refactor the adaptive refinement code to more easily support current and future features.

The core face selection code was extracted and greatly simplified using new tags that had been previously added to support inf-sharp feature recognition.  Even if not using inf-sharp patches, these tags exist and greatly simplify the recognition of regular boundaries.  The selection code for face-varying channels was also extracted into a separate function and the code common to both of them (inf-sharp feature recognition -- as the face-varying faces of interest are all boundaries, it involves little else) was extracted so that both can make use of it.

At a higher level, the FeatureMask used by selection was expanded to include a more comprehensive set of features.  Since the features of interest at a particular level are solely determined by the growing list of AdaptiveOptions, methods to translate user options into features and to reduce that set of features (when using separate isolation levels) were provided.  The revised FeatureMasks are now passed down to face selection functions that operate solely on the feature set specified for a given level and no longer interact with the AdaptiveOptions or any other state of the Refiner in any way.

### DIFF
--- a/opensubdiv/far/topologyRefiner.cpp
+++ b/opensubdiv/far/topologyRefiner.cpp
@@ -239,18 +239,115 @@ TopologyRefiner::RefineUniform(UniformOptions options) {
     assembleFarLevels();
 }
 
-class TopologyRefiner::FeatureMask {
-public:
-    typedef unsigned int int_type;
+//
+//  Internal utility class and function supporting feature adaptive selection of faces...
+//
+namespace internal {
+    //
+    //  FeatureMask is a simple set of bits identifying features to be selected during a level of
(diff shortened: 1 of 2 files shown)


takahito-tejima:
George and I reviewed this request. Thanks Barry!

PR #854, opened 2016-09-28
Added new options for adaptive refinement

barfowl:
This change exposes controls for adaptive refinement for face-varying patches and inf-sharp patches that have been internalized to this point.  It also includes the secondary isolation level to reduce excessive refinement of smooth vs semi-sharp features.


davidgyu:
Looks good to me!

PR #855, opened 2016-09-28
Fixed topology tagging bug affecting face-varying patches

barfowl:
This change corrects a regression that was introduced when face-varying inspection was switched to use the new tags for inf-sharp features, resulting in incorrect refinement for some face-varying patches.  It was most noticeable in the "linear boundaries" case as sharp boundaries were not being correctly identified.


davidgyu:
Thanks for the fix!

PR #867, opened 2016-09-30
Added inverse of Far::PatchParam::Normalize()

barfowl:
Given requests for the inverse of PatchParam::Normalize(), a new Unnormalize() method has been added in place of the two new mapping methods that were recently added.


davidgyu:
Excellent! Thanks!

PR #873, opened 2016-10-01
Fix find_package for zlib which requires all upper case ZLIB

nyue:


davidgyu:
Good catch! I don't think we have an OpenSubdiv CLA from you (OpenSubdiv and USD have separate CLAs). Thanks! -David

nyue:
Hi David,
I have now submitted the OpenSubdiv CLA (individual).
Cheers

davidgyu:
Thanks Nicholas!

davidgyu:
Oops, just realized this PR was to master instead of dev. I will fix that...

PR #890, opened 2016-10-14
Added a few more test shapes and updated viewers

barfowl:
This changes adds a few more test shapes to regression/shapes for face-varying and inf-sharp patches.  Also updated the relevant example viewers to include them.


davidgyu:
Excellent! -David

PR #899, opened 2016-10-21
Fixed glImaging errors w/ ctest

davidgyu:
Fixed some incorrect shader configurations hit
while running the glImaging example through CTest


takahito-tejima:
We reviewed this change together. thanks for fixing!

PR #900, opened 2016-10-21
Templatize far

dcclyde:


jtran56:
Filed as internal issue #151988.

melevine:
How about merging this in? :)

dcclyde:
I've checked that this merged version compiles and that the test I added (derivativeTests) still passes. far_regression passes its tests but hbr_regression currently fails some tests. I haven't checked whether that's my fault.

Do you have an automated linter to check whether I've missed fixing any places where the templatization screwed up your whitespace?

davidgyu:
Thanks again for pushing this forward and continuing to update this pull request!

As you can see, the recent changes we've made to the implementation of internal classes has created a number of conflicts with this pull request.

We've  been working to expose templated precision for the classes and methods needed to take full advantage of double precision for refinement and evaluation. This includes some work beyond the scope of this pull request that also allows internal calculations to be evaluated with greater precision. That work is largely complete and we will be sharing those changes shortly.

barfowl:
Full support for templated precision was finally released in version 3.4 on 6/28/2019, so there will be no need to merge this request.  Please let us know if there are any issues with the revised interface or implementation.  And thanks again for opening this and pushing us to this conclusion.

PR #905, opened 2016-10-27
WIN32 Glew and DLL fixes to CMakeLists.txt

jcowles:
Previously, dynamic targets were disabled on Windows due to configuration
errors in the CMake install targets. In addition, glew was forced to
static linkage on windows, even when dynamic Glew libraries were linked
which would result in linker errors.

In this change, static glew libraries are detected by looking for the "s"
suffix convention on the library (though, maybe this should be handled in
the FindGlew module).

In addition, there is now a fork in the opensubdiv osd_dynamic_{cpu|gpu}
install targets to account for the Windows DLL CMake quirks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -530,10 +530,11 @@ else()
 endif()
 
 if (WIN32)
-    add_definitions(
-        # Link against the static version of GLEW.
-        -DGLEW_STATIC
-    )
+    if ("${GLEW_LIBRARY}" MATCHES "glew32s(d|)")
+        add_definitions(
+	    -DGLEW_STATIC
+	)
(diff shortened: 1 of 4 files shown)


davidgyu:
Hey Jeremy! Thanks again for the PR & CLA

I'm going to cherry pick this but exclude the changes related to osd_dynamic_{cpu,gpu} install and exe linking. Those are the parts of the PR which are causing the AppVeyor failures (which also repro for me locally).

There's more work that we need to do to really support DLL builds, e.g. symbol exports, distinct names for the static and dynamic .libs, etc.

davidgyu:
Closing this in favor of #961

PR #906, opened 2016-10-30
Resolve readme CMake snippet typos

beaverb1ll:
This patch removes some extra `-D `flags that are located in the CMake build snippets. The extra `-D`'s prevent copy and pasting to run the build examples. There was also a trailing white space, so that was also removed.


beaverb1ll:
Oops, this was off of the wrong branch.....

PR #907, opened 2016-10-30
Resolve readme CMake snippet typos

beaverb1ll:
This patch removes some extra `-D` flags that are located in the CMake build snippets. The extra `-D`'s prevent copy and pasting to run the build examples. There was also a trailing white space, so that was also removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Feel free to use it and let us know what you think.
 
 For more details about OpenSubdiv, see [Pixar Graphics Technologies](http://graphics.pixar.com).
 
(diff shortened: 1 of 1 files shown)


davidgyu:
Thanks for the fix! Can you sign and return a CLA so that we can merge your PR?

beaverb1ll:
Not a problem. I sent the CLA on Oct 30 and received a CLA confirmation on Nov 2. If there is something more that is required, just let me know.

davidgyu:
Ah, I see it now (missed earlier). Thanks!

PR #908, opened 2016-11-11
Fixed cmake GLFW X11 link dependencies

davidgyu:
There is more we can clean up here, but this
fixes link issues while building on X11 platforms
when using cmake 3.x


c64kernal:
Looks good, thanks David!

PR #917, opened 2017-01-11
Added option to build using MSVC with static CRT.

davidgyu:
This is a simplified implementation of the work by thomthom to add
support for linking the osd libraries with the static C runtime library
on Windows. We will likely revisit this when we finish adding support
for building the osd libraries as DLLs on Windows.


thomthom:
LGTM

PR #918, opened 2017-01-11
Fixed spelling errors and typos in comments

davidgyu:
This is work by significant-bit and thomthom which
I've cherry-picked and merged to account for recent
changes.


barfowl:
Looks good overall, but one change here reverts a correction to a Doxygen comment made in 3.1...

See line 401 of far/patchTable.h -- use of (u,v) for public methods was made consistent in the comments in 3.1 but the change here reverts it back to (s,t).

davidgyu:
Ah, thanks for the careful review, Barry!

I've fixed that regression and updated the PR

PR #928, opened 2017-01-30
Improved XFB Evaluation of 1st and 2nd derivatives

davidgyu:
Most GL implementations support a maximum of 4 transform
feedback buffer bindings. With the addition of 1st and 2nd
derivative evaluation up to 6 bindings might be required,
i.e. dst, du, dv, duu, duv, dvv.
    
This change extends the GLXFB Evaluator interface to allow
a client to specialize the evaluator when it is known that
(at least) the 1st derivative and 2nd derivative outputs
are interleaved together into shared buffers.
    
When this option is used, the maximum number of transform
feedback buffer bindings can be reduced to 3 instead of 6.

### DIFF
--- a/examples/glEvalLimit/glEvalLimit.cpp
+++ b/examples/glEvalLimit/glEvalLimit.cpp
@@ -507,6 +507,15 @@ private:
     DEVICE_CONTEXT *_deviceContext;
 };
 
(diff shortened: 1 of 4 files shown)


takahito-tejima:
we reviewed this together. thanks!

PR #929, opened 2017-01-31
Fixed Cpu Omp and GLXFB EvalStencils w/ deriv

davidgyu:
Fixed a few bugs uncovered during testing.


takahito-tejima:
looks great, thanks for fixing.

PR #930, opened 2017-04-16
Fixed type issues.

klaussilveira:


barfowl:
Thanks Klaus.  It's not clear what the motivation for these changes is, but there are a couple of issues with the changes that I'll comment on in the code...

examples/common/hdr_reader.cpp
```diff
@@ -81,7 +81,7 @@ unsigned char *loadHdr(const char *filename, HdrInfo *info, bool convertToFloat)
     while(true) {
         if (! fgets(buffer, MAXLINE, fp)) goto error;
         if (buffer[0] == '\n') break;
-        if (buffer[0] == '\r' && buffer[0] == '\n') break;
+        if (buffer[0] == '\r' || buffer[0] == '\n') break;
```
barfowl: With the change to "||" the second condition is not necessary as it always evaluates to false given line 83.  Either the new line 84 makes line 83 redundant, in which case 83 can be removed, or line 84 can be reduced to its first condition.

regression/osd_regression/main.cpp
```diff
@@ -639,17 +639,17 @@ main(int argc, char ** argv) {
...
         printf("Failed to open window.\n");
         glfwTerminate();
         return 1;
     }
     glfwMakeContextCurrent(g_window);
-    
+
 #if defined(OSD_USES_GLEW)
-    if (GLenum r = glewInit() != GLEW_OK) {
+    if (GLenum r = (glewInit() != GLEW_OK)) {
```
barfowl: Adding parenthesis here shouldn't have any effect as "=" is one of the lowest precedence operators.  Both the old and new code are flawed in that the RHS of the assignment evaluates to a boolean value and is then assigned to "GLenum r".  The return code from glewInit() is therefore lost when it is most needed in the error message that follows -- it will always print "error = 1".

Incidentally, it appears this test of the result of glewInit() is coded the same way in all GL examples.

davidgyu:
Hi Klaus,

Thank you for the pull-request. In order to accept your changes we'll need a signed Contributor License Aggrement (CLA) as outlined in: [documentation/contributing](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/master/documentation/contributing.rst).

Also, as Barry referenced, it's best for pull-requests and commits to be for specific changes or fixes described clearly in the comments. We like the guidelines described in: [Pull Request Etiquette](https://gist.github.com/mikepea/863f63d6e37281e329f8).

Thanks!
-David

jtran56:
Filed as internal issue #151989.

barfowl:
Closing this issue.  There was no described motivation for the changes, response to issues with those changes or any response from the contributor (including CLA) in almost three years now.
The flawed reporting of GLEW initialization errors that was touched on here does warrant separate attention across all examples that use it.

PR #932, opened 2017-04-20
Removed shadowing warnings from source for core libraries

barfowl:
The warning flags for GCC do not enable shadowing warnings, which will appear for ICC.  Such warnings were eliminated in the 3.0 source but a few have crept into new code written since.  All of the occurrences are in opensubdiv/far and are corrected here.

I'd prefer we enable -Wshadow for GCC at some point -- even if only for the core library source -- as some of the cases not detected can be pitfalls for someone modifying existing code.

There are a manageable number of cases to deal with in the examples.  Unfortunately, the template-heavy code included from opensubdiv/hbr is riddled with such warnings -- we could eliminate these but since Hbr is being phased out, removing Hbr dependent code from the examples would be preferable.  There are also external header files included that will generate such warnings, particularly Ptex's <Ptexture.h>, but these could be dealt with locally in the few examples that use them.


davidgyu:
This is great! Thanks Barry!

Yes I agree, it would be great to enable -Wshadow (or equivalent for other compilers).

PR #933, opened 2017-05-18
Proposed abstraction of patch building tools into a PatchBuilder class

manuelk:
Proposed abstraction of patch building tools into a PatchBuilder class to be shared between the PatchTableFactory and new topology hashing algorithms.

- add PatchBuilder class
- inherit PatchBuilderContext from PatchBuilder in PatchTableFactory
- move end cap types enum to types.h
- propagate 'fallout' from these changes


manuelk:
Fixing issues - will file a new request...

Issue #922, opened 2017-01-19
Build of osd_static_gpu failed (on Windows 10 using Visual Studio 2013 x64)


huybui:
There is an error in the CMake-generated file (build\CMakeFiles\osd_static_gpu.dir\osd\osd_static_gpu_generated_cudaKernel.cu.obj.Release.cmake)
at line 81:
`set(CUDA_NVCC_COMPILE_DEFINITIONS "**OPENSUBDIV_VERSION_STRING="3.1.1";NOMINMAX;_**USE_MATH_DEFINES;_CRT_SECURE_NO_WARNINGS;OPENSUBDIV_HAS_OPENGL;GLFW_VERSION_3;OSD_USES_GLEW;OPENSUBDIV_HAS_GLSL_TRANSFORM_FEEDBACK;OPENSUBDIV_HAS_GLSL_COMPUTE;OPENSUBDIV_HAS_OPENCL;OPENSUBDIV_HAS_CLEW;OPENSUBDIV_HAS_CUDA;OPENSUBDIV_HAS_PTEX;PTEX_STATIC;OPENSUBDIV_HAS_DX11SDK") # list (needs to be in quotes to handle spaces properly).`

Below is the output of Visual Studio:

`31>  Building NVCC (Device) object opensubdiv/CMakeFiles/osd_static_gpu.dir/osd/Release/osd_static_gpu_generated_cudaKernel.cu.obj
31>  CMake Warning (dev) at osd_static_gpu_generated_cudaKernel.cu.obj.Release.cmake:81:
31>    Syntax Warning in cmake code at column 63
31>  
31>    Argument not separated from preceding token by whitespace.
31>  This warning is for project developers.  Use -Wno-dev to suppress it.
31>  
31>  -- Removing C:/Workspace/OpenSubdiv/build/opensubdiv/CMakeFiles/osd_static_gpu.dir/osd/Release/osd_static_gpu_generated_cudaKernel.cu.obj
31>  
31>  "C:/Program Files/CMake/bin/cmake.exe" -E remove C:/Workspace/OpenSubdiv/build/opensubdiv/CMakeFiles/osd_static_gpu.dir/osd/Release/osd_static_gpu_generated_cudaKernel.cu.obj
31>  
[...]

PR #934, opened 2017-05-22
Addressed issue 992.

carlospzlz:
- Fixed dangling quotes produced by cmake file.
- Added token stringification to convert OPENSUBDIV_VERSION_STRING into
a string.


jtran56:
Filed as internal issue #151990.

davidgyu:
Thanks for looking into this. We're going to close this PR based on our results from recent testing. See notes on issue #922

PR #935, opened 2017-05-23
Patch builder

manuelk:
Merging some refactoring to pave the way for topology hashing implementation (Sig 2016 paper)


jtran56:
Filed as internal issue #151991.

barfowl:
This should now be addressed by the rafactoring that was merged for #966.  The description of that Pull Request provides a good overview.  Some of that work is ongoing and there is room for improvement, so if it doesn't meet your needs, let us know.

Thanks

PR #941, opened 2017-07-14
Fixed mtlPtexViewer build

davidgyu:
Updated mtlPtexViewer to account for recent
API changes to Osd::MTLPatchShaderSource and
also fixed a build error when building with
the use of PtexCache disabled.


c64kernal:
Thank you!

PR #942, opened 2017-07-14
Fixed mtlViewer face-varying evaluation

davidgyu:
Updated the mtlViewer to compute correct offsets
for face-varying param and index buffers when using
the Gregory Basis end cap approximation.


c64kernal:
Very nice find!

PR #945, opened 2017-07-28
Added build options: NO_GLFW NO_GLFW_X11

### LINKED ISSUE #903: X11, GLFW, and OSX oh my
The detection of GLFW has got way too many places to look for macOS, and as a result can be led astray if you have an X11 install. Similarly to USD, OpenSubdiv should have a mechanism to opt out X11 on macOS.

Related, the regressions and examples require GLFW. Unfortunately there is no way to opt out of GLFW, and when you are in an X11 tangle...

You can bypass the regressions and examples like this -

-DNO_OPENGL=1
-DNO_REGRESSION=1

you'll get past the problem, but sadly you will also lose OpenGL everywhere.

I think that there should be a NO_X11 switch, and a NO_GFLW switch.

davidgyu:
This makes it a little easier to eliminate build
dependencies when the interactive examples
are not needed.

Fixes #903


meshula:
Thanks!!

PR #946, opened 2017-09-13
Fixed bug in face-varying patches of uniformly refined Far::PatchTable

barfowl:
This change fixes a bug in face-varying patches for a Far::PatchTable that was constructed from uniform refinement.  The separate PatchParams for the face-varying patches (introduced in 3.1) were not being initialized, so the reparameterizations associated with the depth of the patches were not being computed correctly.


davidgyu:
Thanks for the fix, Barry! This looks great to me!

Issue #947, opened 2017-09-16
GPU patch shaders do not compute correct normals in common degenerate cases


barfowl:
The patch shaders all appear to compute normals as the cross product of dPdu and dPdv and make no attempt to determine a well-defined normal when this cross product becomes zero -- either when one of two the partials is zero or when the two are parallel.

These situations are reasonably common, and the failure of the shaders to compute a well defined normal for them results in undesirable artifacts.  These visual artifacts have also been hindering adoption of new features that provide simpler (fewer) and more accurate patches for the limit surface.

Here is an example of three cases in the glViewer -- first illustrating the cages and patch coloring followed by the same shapes with simple normal shading for clarity: 

![before-patch](https://user-images.githubusercontent.com/4130807/30508533-eee8ef08-9a4d-11e7-80de-46d86452d333.png)

Left is a simple surface of revolution with four quads each with a degenerate edge at the pole.  Center is a simple smooth corner patch with an appropriately smooth boundary curve for the corner.  Right is an example of a smooth corner patch in the interior when infinitely sharp patches are enabled.

![before-normal](https://user-images.githubusercontent.com/4130807/30508534-f754d954-9a4d-11e7-8708-0b49be8a9cb2.png)

I have a fix for this issue that I will be submitting in the near future, but only for the GLSL versions (I'm unable to test others).

I'm opening this issue separately to better illustrate and discuss the problem here -- we may ultimately need multiple pull requests to fully address it.  There may also be some other issues that warrant further discussion, e.g. the choice of tolerance to test for degeneracy, possible change in behavior due to fixing a bug in normal derivative calculation, etc.

Fortunately both the BSpline and Gregory patch shaders make use of Bezier patch evaluation, and rewriting that in a way that efficiently allows detection and resolution of the degeneracy produces the following:

![after-normal](https://user-images.githubusercontent.com/4130807/30508540-1cf8b446-9a4e-11e7-9935-c4b0dd03ffad.png)

This should come at little or no added cost.  Based purely on operation counts, the revised evaluation is comparable to the existing evaluation and the added degeneracy test and determination of the correct normal is marginal.  Testing thus far indicates the revised evaluation may actually be slightly faster, but that testing has been limited.

[...]

PR #948, opened 2017-09-19
Improved GLSL patch shaders to compute normals in common degenerate cases

barfowl:
As noted in Issue #947, GPU shaders do not bother to compute proper unit normal vectors in some common cases where the cross product of dPdu and dPdv is zero.  This change updates the GLSL patch evaluation shaders to do so.  The revised evaluation should not degrade performance and no changes to the public shader API were made.  For those using the shaders to evaluate partial derivatives of the normal vector (dNdu and dNdv), a bug involving the magnitudes of these derivatives was corrected and so may lead to some change in behavior.

The BSpline patch evaluation uses conversion to a Bezier patch and then invokes the Bezier patch evaluation, but the Gregory patch evaluation was unfortunately duplicating the code for Bezier patch evaluation inline.  So the Gregory patch evaluation was first modified to invoke the existing Bezier patch evaluation, which was then revised to affect both.

Regarding changes to the evaluation of the normal partial derivatives, the previous use of the Weingarten equations was replaced as they are not well-defined when the normal degenerates (particularly when dPdu or dPdv is zero).  The Weingarten equations express the normal partials in terms of the position partials, i.e. they express dNdu and dNdv within the tangent plane spanned by dPdu and dPdv, and they are appropriately scaled for a unit normal.  But an additional projection was being applied to the results (lines 1323 and 1324) and while the effect of the second term here is zero (N and dNdu or dNdv being orthogonal) the first term incorrectly scales the results.  (Hopefully we don't have to conditionally preserve the results of the old evaluation within #ifdef.)

For anyone wanting to test this change without moving to 3.3, it should be possible to simply replace osd/glslPatchCommon.glsl with the version here in any version 3.1 or newer.

### DIFF
--- a/opensubdiv/osd/glslPatchCommon.glsl
+++ b/opensubdiv/osd/glslPatchCommon.glsl
@@ -1230,113 +1230,187 @@ OsdEvalPatchBezier(ivec3 patchParam, vec2 UV,
                    out vec3 P, out vec3 dPu, out vec3 dPv,
                    out vec3 N, out vec3 dNu, out vec3 dNv)
 {
-#ifdef OSD_COMPUTE_NORMAL_DERIVATIVES
-    float B[4], D[4], C[4];
-    vec3 BUCP[4] = vec3[4](vec3(0), vec3(0), vec3(0), vec3(0)),
-         DUCP[4] = vec3[4](vec3(0), vec3(0), vec3(0), vec3(0)),
-         CUCP[4] = vec3[4](vec3(0), vec3(0), vec3(0), vec3(0));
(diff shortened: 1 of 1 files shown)


davidgyu:
For anyone wanting to test this on DX or Metal, there is an initial port of these changes to HLSL and Metal (osd/hlslPatchCommon.hlsl and osd/mtllPatchCommon.mtl respectively) in the glsl_patch_normals_testing branch of my forked repository.

jtran56:
Filed as internal issue #151993.

davidgyu:
I'll submit separate pull requests for the HLSL and Metal implementations of this fix.

Issue #737, opened 2015-09-10
vertex indices for Far::PatchTable are non-intuitive for uniformly-refined patches


barfowl:
When constructing a Far::PatchTable from a uniformly refined mesh, a different indexing scheme is used for the control vertices of the patches compared to what is used for adaptive refinement.  This is confusing and warrants attention.

For adaptive patches, vertices for all refined levels are included and indices begin at level 0.  For uniform patches, intermediate levels are excluded and vertices are expected from the last level in addition to those of the cage (level 0).  Using only vertices in the last level makes some sense, but this latter requirement that the cage vertices also be included is unexpected and unnecessary.

The addition to PatchTableFactory::Options is one possibility to improve this.  Three choices that makes sense for patches derived from a uniform refinement are to include all levels (same as adaptive), the last only, or the last level in addition to the cage (a legacy choice that may eventually be removed).

barfowl:
Just drawing attention to this issue as there will likely be upcoming additions to PatchTableFactory::Options to support face-varying patches in 3.1.  So it would be a good time to consider additional Options to address this issue at the time.

manuelk:
IIRC, the 3 options you described above were lost in the transition from OSD 2.x to 3.x - so to me they make sense.

We might need to add some bit-flags to the data structure itself to track which "mode" it has been created with.

In the long run, we will probably want to move on from patch tables, but that's a topic for another discussion...

barfowl:
This issue came up again after helping someone debug some unexpected behavior with face-varying patches (leading to #946).  The issue here is made worse by the fact that vertex/varying patches are not treated the same as face-varying patches when it comes to indices of the patch points and the levels involved.  The vertex/varying patches expect the base level vertices to be include while face-varying patches do not.

I added some improvements in #949, including Doxygen comments for PatchTableFactory::Create().  The options added internal to the source are worth considering for public exposure in a future release that includes API changes (a little simpler than the three options previously noted given the apparent need to treat face-varying separately).

jtran56:
Filed as internal issue #151657.

barfowl:
This has been addressed by #986.  The default behavior remains non-intuitive, but public options are now available to control it.

PR #949, opened 2017-09-21
Improved documentation of vertex indexing scheme for uniform PatchTables

barfowl:
In response to Issue #737 (since it has arisen again), these changes improve documentation of Far::PatchTableFactory::Create() in the hope that its indexing scheme for a set of uniformly refined patches is better understood by those building a Far::PatchTable directly.

Doxygen comments were added the header file and internal options were added to the source so that it is more self-documenting should this arise again.  The internal options may be made public in a future minor release, pending further discussion in #737.

### DIFF
--- a/opensubdiv/far/patchTableFactory.cpp
+++ b/opensubdiv/far/patchTableFactory.cpp
@@ -921,6 +921,13 @@ PatchTableFactory::createUniform(TopologyRefiner const & refiner, Options option
 
     BuilderContext context(refiner, options);
 
+    // Default behavior is to include base level vertices in the patch vertices for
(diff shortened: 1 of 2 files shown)


barfowl:
Wow, that second commit should not have been necessary to fix the failure of the glImaging test in the Travis build, but it did.  The memset(0) call I added back should be redundant given initialization in the loop that follows -- curious as to why execution differs.  I've had no trouble with gcc builds of glImaging locally and that's what failed above.

And this Pull Request was supposed to be a no-brainer...

barfowl:
I discovered a couple of other bugs that can occur with face-varying indices while investigating the original build failure (though none causing the failure).  I'm going close this and resubmit a single commit with those included.

PR #950, opened 2017-09-22
More fixes for face-varying patches in uniform PatchTables

barfowl:
Corrected some minor issues related to indices of face-varying patches in uniform patch tables that may have caused problems in rare circumstances.  

In response to issue #737, also made the indexing discrepancy between vertex/varying patches and face-varying patches clearer with internal options and new Doxygen comments.

### DIFF
--- a/opensubdiv/far/patchTableFactory.cpp
+++ b/opensubdiv/far/patchTableFactory.cpp
@@ -921,6 +921,13 @@ PatchTableFactory::createUniform(TopologyRefiner const & refiner, Options option
 
     BuilderContext context(refiner, options);
 
+    // Default behavior is to include base level vertices in the patch vertices for
+    // vertex and varying patches, but not face-varying.  Consider exposing these as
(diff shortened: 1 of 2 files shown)


barfowl:
I give up...  There's something screwy about the specific glImaging test in the CI build -- it doesn't even require face-varying patches in PatchTableFactory::createUniform() but any change to the stack of that method seems to trigger an std::bad_alloc failure somewhere.  Closing this for now until I have more time to investigate.

davidgyu:
This is a problem with the CI configuration and the glImaging test. Often triggering a re-build will get past the error. There must be some inconsistency in the CI environment and it would be great to get to the bottom of that, but we can also improve the glImaging test to be more robust.

barfowl:
Re-opening this and will see if we can get the glImaging test to pass with a couple of retries...

davidgyu:
Thanks, Barry!

PR #951, opened 2017-09-23
Improved stability of examples/glImaging

davidgyu:
The glImaging test now checks the GL version of the
current context before attempting to run GL XFB or
GLSL Compute kernels.

This example runs as part of the ctest suite during
continuous integration testing and adding these checks
should improve consistency of runs across different
test conditions.


davidgyu:
Well, that was unsatistfying. The first Travis CI run for this PR failed in the same way that it has before while running this test, but re-triggering the CI build passed. Continuing to investigate the failures...

barfowl:
I actually find the failure very satisfying :)

davidgyu:
Indeed. The added commit will disable the glImaging test during Linux CI runs until we get to the bottom of this, so that we don't have to deal with the test failure uncertainty.

PR #954, opened 2017-10-04
Address issue #953

### LINKED ISSUE #953: /Wall + /WX break build on MSVC when using NMake
CMakeLists.txt has the following, starting from line 214:
```
elseif(MSVC)

    # Turn on all warnings
    list(APPEND OSD_COMPILER_FLAGS /Wall)

    list(APPEND OSD_COMPILER_FLAGS
                    /W3     # Use warning level recommended for production purposes.
                    /WX     # Treat all compiler warnings as errors.
```

The code adds `/Wall`, `/W3` and `/WX` to a list of compiler flags. A Visual Studio project, generated by CMake, picks `/W3`, but NMake seems to be using `/Wall`, and this breaks build on the first [info-level](https://stackoverflow.com/questions/4001736/whats-up-with-the-thousands-of-warnings-in-standard-headers-in-msvc-wall) warning.

I'd like to submit a pull request to fix this. What would be the best way to proceed?

1. Just remove `/Wall`, leaving `/W3`.
2. Remove `/WX` too, since it's not future proof in a sense that a new compiler version can introduce new warnings that will prevent builds.

Additionally, I think it's a good idea to simplify lines 128-132 to remove CMake warnings regarding policy [0054](https://cmake.org/cmake/help/latest/policy/CMP0054.html):
```
# Detect Clang (until a cmake version provides built-in variables)
if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
    set(CMAKE_COMPILER_IS_CLANGCC 1)
elseif(CMAKE_CXX_COMPILER_ID MATCHES "Intel")
    set(CMAKE_COMPILER_IS_ICC 1)
endif()
```
The syntax is `if(<variable|string> MATCHES regex)`, which means that we can use variable name instead of its value.

stolyaroleh:
This fixes #953:

* Prevent CMake warnings related to policy [0054](https://cmake.org/cmake/help/latest/policy/CMP0054.html).

* Fix NMake builds by removing `/Wall`, leaving only `/W3` in MSVC compiler flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,9 +125,9 @@ if (CMAKE_COMPILER_IS_GNUCC)
 endif()
 
 # Detect Clang (until a cmake version provides built-in variables)
-if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
     set(CMAKE_COMPILER_IS_CLANGCC 1)
-elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "Intel")
+elseif(CMAKE_CXX_COMPILER_ID MATCHES "Intel")
     set(CMAKE_COMPILER_IS_ICC 1)
 endif()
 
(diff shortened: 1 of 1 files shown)


jtran56:
Filed as internal issue #151996.

davidgyu:
Thanks for this contribution! Fixed by #1318 and #1180

PR #955, opened 2017-10-31
Fixed array index handlings of CpuPatchTable and Level

ryohei-nishimura:
Fixed array index handlings of CpuPatchTable and Level that cause segmentation faluts.

### DIFF
--- a/opensubdiv/osd/cpuPatchTable.cpp
+++ b/opensubdiv/osd/cpuPatchTable.cpp
@@ -97,7 +97,8 @@ CpuPatchTable::CpuPatchTable(const Far::PatchTable *farPatchTable) {
             // face-varying param
             Far::ConstPatchParamArray
                 fvarParam = farPatchTable->GetPatchArrayFVarPatchParams(j, fvc);
-            for (int k = 0; k < numPatches; ++k) {
+            int nPatch = farPatchTable->GetNumPatches(j);
+            for (int k = 0; k < nPatch; ++k) {
                 PatchParam param;
                 //param.patchParam = patchParamTable[patchIndex];
                 param.field0 = fvarParam[k].field0;
--- a/opensubdiv/vtr/level.cpp
+++ b/opensubdiv/vtr/level.cpp
@@ -1599,12 +1599,8 @@ namespace {
             int memberMax = _memberCountPerComp;
             for (int i = 0; i < _compCount; ++i) {
                 int count = _countsAndOffsets[2*i];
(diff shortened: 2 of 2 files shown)


barfowl:
Thanks for the fix.  Could you provide a little more of a description of when this is necessary -- in part so that we can help document this later?

It looks to me that it can only occur when you have a non-manifold mesh with a vertex with no incident edges or faces or an edge with no incident faces.  And it looks like there must be other conditions to trigger the problem (the last component being that bare vertex or edge).  But I am unable to trigger the problem with GCC -- are you building with MSVC?

Thanks

jtran56:
Filed as internal issue #153144.

ryohei-nishimura:
Thank you for your reviewing.

These problems occur at our business project with Visual Studio 2015. And other compilers such as GCC have not been tested yet. We will try a test with MinGW GCC.
The data that trigger the problems are unfortunately confidential. If you would need a precise information about conditions, we could create a sample to replay the problems.

barfowl:
Thanks, I was afraid you might not be able to share the data that triggered it, but that's not unexpected.

Glad to hear the problem was with Visual Studio -- we have had some similar problems with Visual Studio's use of std::vector when we want to retrieve addresses for pointer arithmetic.  I suspect that is the case here as the address ends up being equivalent to dstMembers.end().

Rather than the change you submitted, I recommend the following in place of the "if (count == 0)" condition that was added:

    1603                Index *dstMembers = &dstIndices[0] + _countsAndOffsets[2*i + 1];

The existing block of code already works fine in general when count == 0.  I would prefer not to add additional conditions to this block to slow down the 99.9% of cases where count != 0.  This is part of the more general goal to avoid having the support of non-manifold topology slow down the more common manifold cases.

Please confirm if that change solves your problem, thanks.

ryohei-nishimura:
Thank you, barfowl. Your fix works fine on our Visual Studio environment.
It has been pushed into this pull request.

davidgyu:
We will need a signed CLA to proceed with merging this change: [Contributing](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/master/documentation/contributing.rst)

Thanks!
-David

ryohei-nishimura:
Now I am asking my company's permission to sign the CLA. Wait for a little more days please.

barfowl:
Closing this as the issues raised here have since been addressed in #987 and #1001.  Thanks again for raising and helping to address them.

PR #957, opened 2017-11-21
Fixed integer overflow bugs for large meshes in PatchTable factories

barfowl:
Fixed a couple of bugs reserving memory during creation of Far::PatchTables for large data sets.  Estimated upper bounds use the product of two sizes declared as "int"  which overflows and results in negative values passed to vector::reserve() (and so cast to "size_t" and crashing) instead of the smaller fixed positive size that was intended in such cases.


jtran56:
Filed as internal issue #154146.

davidgyu:
Thanks, Barry!

PR #959, opened 2017-12-12
HLSL patch shader changes for degenerate normals

davidgyu:
Updated HLSL patch shaders to resolve degenerate normals
to address #947. This fix was ported from the GLSL patch
shader source.

Also, added missing inf sharp test cases to dxViewer.


jtran56:
Filed as internal issue #155001.

PR #960, opened 2017-12-12
Metal patch shader changes for degenerate normals

davidgyu:
Updated Metal patch shaders to resolve degenerate normals
to address #947. This fix was ported from the GLSL patch
shader source.

Also, added missing inf sharp test cases to mtlViewer.


jtran56:
Filed as internal issue #155002.

barfowl:
Thanks David (and sorry for the accidental closure and re-opening)

PR #961, opened 2017-12-16
CUDA 9 compatible gpu-architecture default

doctaweeks:


jtran56:
Filed as internal issue #155179.

davidgyu:
We've implemented a simpler fix by only setting the gpu architecture here for older (now obsolete) versions of CUDA and it's still possible to use OSD_CUDA_NVCC_FLAGS for more control.

PR #966, opened 2018-05-14
Performance improvement and refactoring of Far::PatchTable construction

barfowl:
This is a significant set of changes affecting the internal construction of Far::PatchTable, but not the nature of the PatchTable itself.  Construction time is greatly improved for meshes that are highly irregular (2-5x).  More importantly, the refactoring involved simplifies the future construction of PatchTables from other subdivision schemes (e.g. Loop) and provides information about patches independent of the PatchTable itself or its factory.

Highlights include:

  * Far::PatchTableFactory now completely defers the construction of a PatchTable to the new, locally defined PatchTableBuilder class.  This combines and replaces the factory class' private methods and the separate, locally declared context class.

  * PatchTableBuilder makes use of the new PatchBuilder class and, indirectly, its subclasses for each subdivision scheme.  PatchBuilder provides information about patches independent of higher level constructs like the PatchTable itself or StencilTables.  This allows the PatchTableBuilder to assemble a PatchTable in more general terms -- independent of the type of subdivision scheme or patches involved.

  * Improvements to the low-level computation of points of irregular patches will result in slight numerical differences.  The formulae used did not specialize the common case of regular topology at three of the four patch corners, so many near-zero terms would result where exact values of zero could be used.  New results are therefore more accurate, require fewer local-point stencils (generally 25% fewer) resulting in slightly lower memory usage and faster computation of local patch points.

All new classes are internal and no public interfaces should be modified here.  All changes to public header files should just be the removal of obsolete declarations previously required for protected/private needs.


davidgyu:
Thanks, Barry!

In addition to the local tests in this repository, we've also done some testing of this change using our studio test suites and the results have been clean (taking into account the slight numerical differences mentioned above).

-David

PR #970, opened 2018-07-03
Fix compilation error when using CUDA Toolkit 9

Nazg-Gul:
Compute capabilities 2.0 and 2.1 (Fermi cards) were removed from
CUDA Toolkit 9. Build CUDA kernels of compute capability 3.0 when
CUDA Toolkit 9 is used.


Nazg-Gul:
Oh, just noticed #961 . Sorry for he noise.

PR #971, opened 2018-07-03
Fix wrong argument order for memset

Nazg-Gul:


barfowl:
Thanks for catching that -- of all the silly mistakes I make over the years that slip by the compiler, that's a new one.

davidgyu:
Updated this Pull Request to merge into dev instead of master

PR #972, opened 2018-07-03
Fixed regression in PatchTable construction with face-varying patches

barfowl:
A regression recently introduced causes construction of non-linear face-varying patches to crash when no adaptive refinement was required.  The error in the estimate of the StencilTable size for the face-varying patches has been corrected, and the catmark_torus shape has had its UVs modified to cover cases like this in future.


davidgyu:
Thanks, Barry!

PR #973, opened 2018-07-04
Fix stencil viewer

Nazg-Gul:
Fixed unintended implicit fall-through in checkbox update callback.


jtran56:
Filed as internal issue #162591.

PR #975, opened 2018-07-14
Fix uninitialized stack memory used in topology refiner

Nazg-Gul:
The issue is that FeatureMask structure has size of an integer, but only
first 11 bytes are initialized, the rest is left non-initialized to a
stack memory of RefineAdaptive(). This will cause issues with IsEmpty()
which check whether any of integer's bits are set, effectively causing
jumps dependent on a non-initialized memory.

### DIFF
--- a/opensubdiv/far/topologyRefiner.cpp
+++ b/opensubdiv/far/topologyRefiner.cpp
@@ -263,7 +263,7 @@ namespace internal {
         bool IsEmpty() const { return *((int_type*)this) == 0; }
 
         FeatureMask() { Clear(); }
-        FeatureMask(Options const & options, Sdc::SchemeType sType) { InitializeFeatures(options, sType); }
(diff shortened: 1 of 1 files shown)


barfowl:
Thanks, this has been creating noise in valgrind since 3.1 and am glad to see that cleared up.

While this change does correct a potential error, the way the bits of the FeatureMask are currently used, such an error can never occur.  Currently some feature bits will always be set on non-default construction and it is not possible to unset all of those bits.  So IsEmpty() should never succeed and the presence of any uninitialized bits elsewhere in the 32-bit word does not affect that outcome.  That will change in future when conditions for adaptive refinement of other schemes are enabled and more bits will be explicitly unset.

Given our recent focus on regressions due to some other changes, I just want to make it clear
to others that this is not another such regression.

PR #976, opened 2018-07-18
Fixed regression in PatchTable construction with varying patches

barfowl:
Recent refactoring of PatchTable construction (#966) introduced a regression with the construction of patches for Varying data.  The bilinear patches would be generated in an incorrect order when more than one array of patches was being constructed, i.e. Gregory patches were used for the irregular patches.  When both regular and irregular patches were represented with BSplines, use of a single patch array avoided the problem.


davidgyu:
Thanks, Barry!

PR #977, opened 2018-07-23
Internal improvements to support double precision

barfowl:
This set of changes improves some of the low-level numerical calculations in Sdc and Far to expand the use of templates in support of double precision.  No public interfaces are affected at this time.

The Sdc schemes were updated to correct some oversights in their support of precision templates (retention of full precision with literal constants, use of std math functions, etc).  Similar computations in Far::CatmarkPatchBuilder were reformatted to support templates for precision.  The Far::PatchBuilder and its usage was extended to make greater use of the SparseMatrix template, and all patch basis evaluation was converted to templates.


jtran56:
Filed as internal issue #163150.

davidgyu:
Looks great! Thanks, Barry!

PR #978, opened 2018-07-27
Superficial header reordering in far/catmarkPatchBuilder.cpp

barfowl:
This is a trivial superficial change that was meant to be included in a recent pull request (#977) -- just reordering the header files in a single source file to conform to established convention.


davidgyu:
Great! Thanks, Barry!

PR #980, opened 2018-08-02
Extend the public interface of Far to support double precision

barfowl:
This set of changes extends the public interface of classes within Far to support double precision.  It achieves the goals of #900 while impacting fewer public classes and preserving the existing public interface.  The changes are applied in two groups/commits.

The first impacts classes that are topological in nature and whose interface only marginally involves floating point values, e.g. Far::PatchTable, Far::PatchParam, etc.  Public methods of these classes that involve floating point values are extended to template methods.  Additional overloads are provided for compatibility in cases where existing usage creates template parameter resolution issues.

The second impacts classes whose primary function is to manage or compute floating point values, e.g. the Stencil classes (along with their associated Tables and Factories) and the Far::PrimvarRefiner.  These have been converted to class templates.  To preserve compatibility, the new templates are given a suffix to distinguish them from the existing class names, and the existing classes have been redefined as light-weight wrappers derived from the new templates (as typedef alone is insufficient).

A third commit modifies a couple of the examples and tutorials to illustrate and exercise the new public templates.


davidgyu:
Thank you, Barry!

PR #981, opened 2018-08-06
Fix minor oversights in recent changes supporting double precision

barfowl:
These changes correct a couple of oversights in the many changes that have been pulled recently in support of double precision in Far:

- fixed the LimitStencilTableFactory wrapper class to accept a StencilTable in its create method instead of a LimitStencilTable
- promoted declaration of a set of constants in CatmarkPatchBuilder to retain double precision


davidgyu:
Thanks!

PR #982, opened 2018-08-06
fix Far::StencilTable append when base StencilTable empty

barfowl:
This change extends Far::StencilTableFactory::AppendLocalPointStencilTable() to support the case when the base StencilTable is empty or un-specified (passing an empty base StencilTable will currently crash).  This case can occur when no adaptive refinement is applied but local points are present for any patch conversions.


davidgyu:
Thank you again!

PR #983, opened 2018-08-06
Revisions to the introductory documentation page on "Subdivision Surfaces"

barfowl:
These changes affect the introductory documentation section titled "Subdivision Surfaces".  Based on issues and questions raised by clients and users in recent years, the focus of the introduction was shifted to emphasize subdivision surfaces as a surface primitive in contrast to a polygonal smoothing operation that many consider it to be.  Subdivision surfaces are defined fundamentally as piecewise parametric surfaces over meshes of arbitrary topology.  The underlying concepts and implications of this definition are described in greater detail, followed by the specifics of OpenSubdiv from the original page.

Further revisions are expected as we gather feedback.  We also observed that previous revisions to the same page had left some unresolved links, and these changes create a few more.  We'll be dealing with all unresolved links at a later date.


davidgyu:
Great!

PR #986, opened 2018-08-27
Added flags to Far::PatchTableFactory::Options to address open issues

barfowl:
Two sets of new flags are added here to Far::PatchTableFactory::Options to address a couple of identified issues.  The code supporting these new options has long been supported internally, so the code changes do nothing more than expose them in the public interface.

The first set addresses indexing consistency issues with Uniform PatchTables raised in #737.  The second set allows the generation of Varying patches and/or their local point StencilTables to be suppressed.  A tutorial was updated to illustrate use of the latter, but there is no tutorial that constructs a uniform PatchTable for the latter.


barfowl:
Looks like the Travis CI builds are failing over a configuration issue:  "Xcursor library not found - required for GLFW"...

jtran56:
Filed as internal issue #OSD-229.

PR #987, opened 2018-08-27
Fix rare non-manifold topology indexing issue in Vtr::Level

barfowl:
This change fixes an indexing error that can occur in a rare non-manifold case but only when runtime bounds-checking is enabled for std::vector (originally identified in #955).  The case occurs when one or more of the last edge or vertex has 0 incident faces or edges.  Address arithmetic is now used to determine the end of the array to avoid the bounds-check errors triggered by use of vector[].


jtran56:
Filed as internal issue #OSD-228.

PR #988, opened 2018-08-28
Minor improvements to internal Far::SparseMatrix class

barfowl:
This change fixes a bug in SparseMatrix::Copy() and adds a new Swap() method.

The Copy() method is not used internally yet (it will actually fail to compile once instantiated) so this is not correcting any existing errant behavior -- I just want to correct this as soon as possible before anyone else runs into it.


jtran56:
Filed as internal issue #OSD-230.

PR #990, opened 2018-09-11
Minor extensions to Far classes to help manage large PatchTables

barfowl:
This set of changes includes several minor extensions to Far classes (typically overloading a method) to help deal with PatchTables for really large meshes.  The time and space required to build a single large PatchTable for evaluation or tessellation can be objectionable, and that construction cannot be threaded.  Each of the changes here is relatively minor, but by dealing with arbitrary pieces of the mesh and a common base level more easily, their combined use provides a lot of flexibility in dealing with large meshes

The changes include:

- overloading PatchTable construction to restrict the patches constructed to a subset of base faces
- overloading adaptive refinement to restrict refinement to a subset of base faces
- overloading TopologyRefiner construction allowing a new instance to be created from and share the base level of an existing instance
- overloading PatchTable local point compute methods to allow base and refined primvar buffers to be separate

There are multiple ways to use these single- or multi-threaded to avoid the construction of a single huge PatchTable.  A new tutorial (far/tutorial_9) was added with a simple example of their usage.

### DIFF
--- a/opensubdiv/far/patchTable.h
+++ b/opensubdiv/far/patchTable.h
@@ -172,6 +172,10 @@ public:
     template <class T> void
     ComputeLocalPointValues(T const *src, T *dst) const;
 
+    template <class T> void
+    ComputeLocalPointValues(T const *srcBase, int numBase,
+                            T const *srcRefined, T *dst) const;
+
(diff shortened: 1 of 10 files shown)


davidgyu:
Thanks, Barry!

PR #991, opened 2018-09-13
Simplify construction of Gregory patches with val-2 interior corners

barfowl:
The previous refactoring of Gregory patch construction (#966) replaced the inefficient accumulation of coefficients with direct assignment -- in all cases but one.  Since the 1-ring of a corner adjacent to a valence-2 interior corner actually overlaps itself, the accumulation was preserved in this case.  Support for this ended up littering the code with conditions specific to the valence-2 case -- complicating what should have been relatively simple assignment in the general case.

This change removes most of the invasive member variables and conditions associated with this case by allowing duplicate entries for the 1-rings to exist for a point, i.e. a row in the conversion matrix, but then accumulating them as a post-process only when a valence-2 interior vertex is present.  The same post-process can also be used for Gregory triangles, along with a similarly simplified general case.

Aside from removing valence-2 conditions from the construction logic, this also allowed a lot of baggage associated with the accumulation of coefficients to be removed.  Two separate commits are submitted that reflect these two stages as the combined diffs are much harder to sift through compared to the two individually (the first stage affects the logic and functionality, the second is essentially superficial).

### DIFF
--- a/opensubdiv/far/catmarkPatchBuilder.cpp
+++ b/opensubdiv/far/catmarkPatchBuilder.cpp
@@ -310,68 +310,44 @@ CatmarkLimits<REAL>::ComputeBoundaryPointWeights(int valence, int faceInRing,
 }
 
 //
-//  SparseMatrixPoint
+//  SparseMatrixRow
 //
 //  This is a utility class representing a row of a SparseMatrix -- which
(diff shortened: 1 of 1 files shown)


davidgyu:
This worked out really well! Thanks, Barry!

Issue #992, opened 2018-09-15
Dev Branch break existing code relating to LimitStencils


oracle3001:
This code was valid until the dev branch.

```
auto g_controlStencils = LimitStencilTableFactory::Create(*_refiner, locs);

	//Evaluate the surface with parametric coordinates
	int n_limit_stencils = g_controlStencils->GetNumStencils();
	for (int i = 0; i < n_limit_stencils; ++i)
	{
		LimitStencil limit_st = g_controlStencils->GetLimitStencil(i); 
                unsigned int size_st = limit_st.GetSize();
.....
```

error C2247: 'OpenSubdiv::v3_3_3::Far::StencilReal<float>::GetSize' not accessible because 'OpenSubdiv::v3_3_3::Far::LimitStencil' uses 'private' to inherit from 'OpenSubdiv::v3_3_3::Far::LimitStencilReal<float>'

davidgyu:
Thanks for filing. Yes, we will fix this oversight ASAP. Thanks!

jtran56:
Filed as internal issue #OSD-240.

davidgyu:
Fixed by #993

PR #993, opened 2018-09-17
Fixed oversight deriving Far::LimitStencil from new template class

barfowl:
As raised in Issue #992, this change fixes the omission of the "public" keyword when deriving Far::LimitStencil from the new template class Far::LimitStencilReal<float>.


davidgyu:
Fixes #992

PR #998, opened 2018-09-25
Remove Catmark-only restriction on adaptive refinement in Far

barfowl:
This change removes the restriction limiting adaptive refinement to the Catmark scheme -- a prerequisite to building patches for Bilinear and Loop.

Adaptive refinement has generally worked for triangles since 3.0, but a couple of subtle issues remained.  The change here mainly deals with forced isolation of boundary features that has been present for quads since 2.x.  For triangles, a single level of refinement does not isolate boundary features -- refinement of a regular corner triangle with three boundary vertices produces a new child face at the corner that also has three boundary vertices.  We don't want such regular faces refined to the maximum depth, and there is no gain in refining at least once, so the adaptive analysis does not force boundary isolation on regular triangles as it does for quads.

Ultimately we do want to eliminate the need for boundary isolation for quads, but that requires elimination of any downstream dependencies and more of a concerted effort.


barfowl:
Opened on master instead of dev -- closing and will re-open.

Issue #767, opened 2015-12-14
Include directories convention and consistency ?


nyue:
Hi,

  In glEvalLimit.cpp, the following convention is used throughout
# include <osd/<blah>.h>

  In far_tutorial_0.cpp, the following convention is used throughout (NOTE additional "opensubdiv" )
# include <opensubdiv/far/<blah>.h>

  What is the OpenSubdiv convention ? Will that be documented for everyone to follow ?

  With a documented convention, we can at least have a frame of reference for cleanup and pull request.

Cheers

jcowles:
Hi Nicholas, yes, this is kind of a mess currently. For examples and tutorials, I would prefer the include paths to match as closely to external client code as possible.

That way, if you're using this as a means of learning the API, you could copy and paste the includes and have some hope of it "just working".

After the new year, we'll be doing another push to release 3.1, so hopefully we can fit in some example cleanup and get the coding convention into public docs.

If you get there first, that would be greatly appreciated! However, please let us know your plans before doing any heavy lifting, to make sure we're all on the same page.

salvager2014:
i know this but you must do this. 发自网易邮箱大师 On 2015-12-14 15:37 , Jeremy Cowles Wrote: Hi Nicholas, yes, this is kind of a mess currently. For examples and tutorials, I would prefer the include paths to match as closely to external client code as possible. That way, if you're using this as a means of learning the API, you could copy and paste the includes and have some hope of it "just working". After the new year, we'll be doing another push to release 3.1, so hopefully we can fit in some example cleanup and get the coding convention into public docs. If you get there first, that would be greatly appreciated! However, please let us know your plans before doing any heavy lifting, to make sure we're all on the same page. — Reply to this email directly or view it on GitHub.

manuelk:
Along with include paths, there are spots in the examples that rely on CMake build knowledge that generally makes transporting this code outside of the OSD build more difficult than it should be (something i have had to do recently...). 

Eventually it would be nice to treat the example code as external apps entirely with only the information that a FindOSD.cmake module would give them (I have one of those btw, it just needs a bit of cleanup). 

This would be a fair amount of work within the current build code and it's not exactly a priority...

jtran56:
Filed as internal issue #151663.

barfowl:
This has been addressed in the dev branch with the merger of #1005 and #1051 and will be included in the next release.

PR #1005, opened 2018-10-11
Step towards consistency of header file inclusion in local source

barfowl:
These changes are a step towards consistency of header file inclusion in source files for the tutorials, regressions, examples, etc., as noted in #767 and elsewhere.  The convention is to include header files the same as external client code, e.g.  as <opensubdiv/far/patchTable.h>.

The default OPENSUBDIV_INCLUDE_DIR definition has been modified to now exclude the ./opensubdiv subdirectory.  All CMakeLists files and tutorial, regression and common source -- everything except the individual examples -- have been updated to enforce the convention.

Source for the examples remains unchanged and each of their CMakeLists have been updated to extend the include path to keep them building in their current state.  So changes required to make each example conform to the convention are now local to that example.  There is ongoing work planned for many of the examples in the near future, and they will be updated on an individual basis.


davidgyu:
Thanks, Barry!

PR #1007, opened 2018-10-15
Populate Far::PatchTable with triangular patches for Loop scheme

barfowl:
This set of changes populates the Far::PatchTable with triangular patches representing the limit surface of a Loop subdivision mesh.

PatchTable construction was previously factored to assemble the tables in more general terms, so no changes to the factory were necessary here.  The internal PatchBuilder base class and its LoopPatchBuilder subclass are extended here to support triangular patches.  The base class handles the topological inspection and extraction of regular patches while the subclass handles the conversion of irregular patches to the supported triangular patch types:  linear, Gregory and the regular Box-spline.

This completes support for Loop limit patches in Far, so tessellation of the limit surface for Loop is now possible via the PatchTable.  The various Osd evaluators and backends do require extension and/or the removal of Catmark/quad specific logic to varying degrees before Loop patches are fully supported.  That work is in progress.
~

### DIFF
--- a/opensubdiv/far/loopPatchBuilder.cpp
+++ b/opensubdiv/far/loopPatchBuilder.cpp
@@ -23,31 +23,1614 @@
 //
 
 #include "../far/loopPatchBuilder.h"
+#include "../vtr/stackBuffer.h"
+#include "../sdc/loopScheme.h"
 
 #include <cassert>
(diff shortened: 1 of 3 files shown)


davidgyu:
Excellent!

PR #1010, opened 2018-10-19
Fix initialization of PatchTable's varying patch descriptor

barfowl:
The varying descriptor of Far::PatchTable is used elsewhere to determine if the table contains triangular or quad patches, but that descriptor is not being initialized if generating varying patches is suppressed, e.g. in far/tutorial_9.  This change initializes the descriptor on construction regardless of whether varying patches are generated.  (Meanwhile alternatives to the use of the varying descriptor are under consideration.)


davidgyu:
Thanks, Barry!

PR #1013, opened 2018-11-01
Improve patch construction around non-manifold vertices to close gaps

barfowl:
Patches around non-manifold vertices have historically been approximated with regular patches, which leaves gaps when an irregular patch is warranted (the catmark_fan shape has illustrated this in the viewer examples).  This change constructs irregular patches where necessary to eliminate gaps.  The faces around a non-manifold vertex are now partitioned into groups delimited by boundary, non-manifold or inf-sharp edges, with each group containing a continuous set of patches.


davidgyu:
This is really great! Thanks, Barry!

PR #1014, opened 2018-11-06
Improved Far::PatchBuilder's determination of patch properties

barfowl:
These changes improve the Far::PatchBuilder's inspection and classification of patch properties to better deal with features that are not isolated.  Previous assumptions that features were isolated via adaptive refinement are not met in the case of Catmark subdivision at level 1 when non-quad faces are present in the base level, and logic to deal with this case was scattered in several places.

The initial test to determine if a patch is regular was re-organized (and better documented) to emphasize whether irregular features are isolated and to deal with one or all accordingly.  Greater use of bit masks is also applied here and elsewhere to simplify dealing with the public options that need to be supported while more efficient acceptance/rejection tests were also added to handle common cases.

The results should be free of assumptions regarding feature isolation of quads (either irregularities or boundaries) and so better able to deal with patches at the base level.


davidgyu:
Thanks, Barry!

PR #1020, opened 2018-11-12
Improve Far::PatchMap for sparse and triangular patch tables

barfowl:
These changes address two limitations in the Far::PatchMap.

For the first, the use of PatchParam bits in constructing the quadtree for rotated triangular patches turned out not to work for all cases of adaptive refinement (e.g. when features were present along edges rather than vertices), so it was replaced with a construction strategy that more closely matches the search.  This lead to some refactoring of the PatchMap initialization to support the differing and common needs of quad and triangular construction.

The second limitation was the treatment of the indices of faces from which the patches in the given PatchTable originated.  Since the PatchTable does not retain the topology of the mesh from which patches were constructed, the PatchMap tried to infer the set of originating faces by identifying the maximum face index of the patches.  But if the last faces in the mesh were tagged as holes and no patches generated, queries related to these last faces would not be correctly detected as intended (child nodes would be present and set where unset root nodes of the face should be).  And with sparse PatchTables becoming more common, the handling of queries for faces not in the map becomes even more unreliable.

So the PatchMap now identifies both the minimum and maximum faces from which the patches originated.  Allocation of quadtree root nodes for the faces now only allocates the detected subrange (which may be a small fraction of a much larger mesh) and searches quickly reject faces outside this supported range.  Any faces within the subrange that do not have corresponding patches will be correctly detected and queries fail appropriately.

### DIFF
--- a/opensubdiv/far/patchMap.cpp
+++ b/opensubdiv/far/patchMap.cpp
@@ -31,144 +31,175 @@ namespace OPENSUBDIV_VERSION {
 
 namespace Far {
 
-// Constructor
-PatchMap::PatchMap( PatchTable const & patchTable ) {
-    initialize( patchTable );
-}
+//
+//  Inline quadtree assembly methods used by the constructor:
+//
 
-// sets all the children to point to the patch of index patchIdx
-void
-PatchMap::QuadNode::SetChild(int patchIdx) {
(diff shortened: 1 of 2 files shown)


davidgyu:
Thanks, Barry! These improvements to the implementation are very nice!

PR #1033, opened 2018-12-14
Assign Far::PatchParam's regular bit consistent with Osd::PatchArray

barfowl:
With patch evaluation in both Far and Osd relying more on the "IsRegular" bit of Far::PatchParam, the bits of the PatchParams for the regular and irregular arrays in a PatchTable have been updated to reflect that regular/irregular status.  The PatchBuilder method most often used to initialize the PatchParam of each patch was extended to optionally specify the value of the regular bit and its usage updated to do so.


davidgyu:
Thanks, Barry!

PR #1034, opened 2018-12-20
Fixed triangular patches at interior infinitely sharp edges

barfowl:
This change fixes a regression for triangular patches that cropped up when simplifying the topological analysis of patches.  When determining the boundary mask of a triangular patch whose apex lies on an interior infinitely sharp edge, the bit indicating the vertex has incident inf-sharp edges should be used rather than the bit indicating the vertex itself is inf-sharp.


davidgyu:
Good catch! Thanks, Barry!

PR #1036, opened 2019-01-04
Fix construction of face-varying patches around non-manifold vertices

barfowl:
With the recent removal of the restriction that patches around non-manifold vertices be regular, face-varying patches whose topology does not match warrants a little more care.  Regardless of the face-varying topology, the fact that the vertex is non-manifold makes is necessary to perform a non-manifold search for the span containing the face of the patch.  So the initialization of the face-varying tags is updated here to ensure that the non-manifold bit of the vertex is always included to trigger the appropriate search.


davidgyu:
Looks good. Thanks, Barry!

PR #1037, opened 2019-01-04
Simplify the private PatchTable::StencilTableHandler class

barfowl:
This is purely an aesthetic change to improve some potentially objectionable implementation details in the PatchTable and doesn't impact functionality or efficiency.

The private PatchTable::StencilTableHandler class was introduced when support for double precision was added to manage member variables for stencil tables of conditional type.  Aside from managing the pointer assignment, this class grew to include some convenience methods that made some implementation details easier, but that obscured the main purpose of the class and created a certain amount of confusion as to its role relative to stencil tables in general.

The change here replaces StencilTableHandler with a simpler StencilTablePtr class whose sole purpose is to retain the stencil table pointer of the conditional type.  It is not a smart pointer that manages the lifetime of the object and includes no methods other than get/set.  This pushes the burden on the PatchTable and PatchTableBuilder classes to deal with the conditional type more directly, but that is generally straight forward.  The addition of new templated initialization methods in the PatchTableBuilder actually simplifies its usage.

### DIFF
--- a/opensubdiv/far/patchTable.cpp
+++ b/opensubdiv/far/patchTable.cpp
@@ -66,25 +66,59 @@ PatchTable::PatchTable(PatchTable const & src) :
     _varyingPrecisionIsDouble(src._varyingPrecisionIsDouble),
     _faceVaryingPrecisionIsDouble(src._faceVaryingPrecisionIsDouble) {
 
-    if (src._localPointStencils.IsSet()) {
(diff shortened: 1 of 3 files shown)


davidgyu:
Thanks, Barry! This looks great!

barfowl:
I just added a commit to remove that now redundant condition you spotted, David -- thanks for catching that.

PR #1053, opened 2019-02-04
Fixed Far::PatchDescriptor::IsAdaptive() for legacy Gregory

barfowl:
The IsAdaptive() test was previously modified to use the number of control points to detect non-linear patches requiring tessellation, but legacy Gregory patches are of size 4.  So the patch type must ultimately be inspected (with or without the patch sizes).


jtran56:
Filed as internal issue #OSD-256.

PR #1054, opened 2019-02-04
Fix cracks between triangular Gregory patches at inf-sharp darts

barfowl:
This change fixes slight cracks between the boundaries of triangular patches on either side of an inf-sharp dart edge.  The logic to compute the quartic point on the boundary between the two edge-points needed to be adjusted to consider the presence of darts (even though both corners may be regular) and not use the trivial assignment in that case.


jtran56:
Filed as internal issue #OSD-257.

PR #1057, opened 2019-02-07
Correction to minor changes to BoxSplineTri boundary computation

barfowl:
Correcting the logic introduced recently in #1055 for patches with two boundary edges.


davidgyu:
Thanks, Barry!

PR #1060, opened 2019-02-12
Minor improvement to degenerate normals in GLSL Bezier triangle

barfowl:
The 1st order Taylor approximation used to resolve degenerate normals in the GLSL Bezier triangle shader was copied from its use with quads.  It approximates in the direction of the patch interior, but detecting the upper U and V patch extremities of the unit square for quads is not sufficient for triangles -- the diagonal boundary (where u + v == 1) also needs to be considered and the derivative terms negated accordingly.


davidgyu:
Looks good. Thanks!

PR #1065, opened 2019-02-28
Minor modification to the contribution docs.

c64kernal:
- Removed explicit username and replaced it with generic example username.


jilliene:
Filed as internal issue #OSD-260

PR #1073, opened 2019-03-05
Fix build script on Linux when not building ptex

c64kernal:
The build script assumed that we always had ZLIB, but we only add it when we build ptex.  Instead of removing it on Linux, we now only add it when we need to on non-Linux platforms.


jilliene:
Filed as internal issue #OSD-274

PR #1074, opened 2019-03-09
Turn on C++11 in clang when building with Metal

c64kernal:
The example code that uses Metal forces C++11 features, therefore we have to turn it on clang so that we can enable the building of compatible code that can be used by Metal and GL at the same time.


jilliene:
Filed as internal issue #OSD-275

PR #1075, opened 2019-03-09
Fixes to the build script for macOS

c64kernal:
- Default the generator to Xcode on macOS
- Fix up how we pass in number of jobs so it works with Xcode


jilliene:
Filed as internal issue #OSD-276

PR #1077, opened 2019-03-10
Fix build of Metal examples when also building GL

c64kernal:
This fixes the build when trying to build both the Metal examples and the GL examples in the same build. The change is an unfortunate consequence of how dependencies are bundled together in examples/common and a future task is needed to split up these dependencies.  The same workaround exists in the DirectX examples as well and would also benefit from splitting up the dependencies.


jilliene:
Filed as internal issue #OSD-278

PR #1078, opened 2019-03-10
Mostly cosmetic clean up to the Metal example apps

c64kernal:
- Renamed the default "NewApplication" in menus to the real name of the app
- Fixed copyright information in the "About" dialog.


jilliene:
Filed as internal issue #OSD-279

PR #1082, opened 2019-03-22
Pruned the overly long list of shapes in example viewers

barfowl:
The fixed list of shapes in the example viewers has grown much longer recently with the addition of new shapes for Loop and Bilinear schemes -- to the extent that the last shapes on the list often do not fit within typical screen sizes.

This change removes redundant, trivial or otherwise poor shapes from the viewer lists so that a reasonably complete set fits in the default window size.  The lists were also copied between the glViewer, dxViewer and mtlViewer for consistency.


c64kernal:
Thanks Barry!

PR #1083, opened 2019-03-23
Reorganized Far tutorials into groups

barfowl:
This change renames the sequential list of Far tutorials from tutorial_0 through tutorial_10 to tutorial_X_Y where X is an integer associated with a group of tutorials sharing a common theme and Y is an integer identifying a tutorial within that group.  While the source directories are named with a X_Y suffix, the tutorials are more commonly referred to as X.Y.

The main documentation page for the tutorials was updated to reflect the new organization.  The two-column tables containing the old tutorials were replaced with a simple hierarchy of subsections -- each group defining its own subsection named according to the functionality it is intended to illustrate.  Other references in the documentation were updated for the new numbering.


c64kernal:
This is great, thanks Barry!

PR #1087, opened 2019-04-09
Factored viewer args parsing for into common place

c64kernal:
Started refactoring how we parse arguments for sample viewer applications into a common place.  This will make it easier to keep args consistent between applications.


barfowl:
George, did you see the build failure on this request?

Looks like a couple of missing header files in argOptions.cpp (<cstdio> etc) that GCC is more picky about in this environment...

c64kernal:
Thanks @barfowl!  Should hopefully be fixed with the latest commit.  Also tested it on clang.

PR #1088, opened 2019-04-09
Added hi-res image files for OpenSubdiv logo to documentation/images

barfowl:
Added new hi-res image files for the OpenSubdiv logo to the repository.  The only image file currently present is around 50x200 pixels and text looks ragged when scaled up in other contexts.  Two versions of the logo are added here:  one with lighter gray text for better contrast with dark backgrounds (matching the lo-res version in current use) and another with darker text for better contrast with light backgrounds.


c64kernal:
Thanks @barfowl

PR #1091, opened 2019-04-12
Minor documentation changes to preserve existing external links

barfowl:
Recent changes to the documentation have been made consistent in terms of internal links, but changes to subsections and their titles have broken pre-existing external links to those sections.  These changes restore some subsection titles and rename explicit internal targets to preserve those external links (which include several in the USD code base and examples).


c64kernal:
Thanks @barfowl

PR #1094, opened 2019-04-23
Moved definition of struct ShapeDesc into same header file as Shape

barfowl:
These changes affect common utilities used by examples and regressions and not the library itself.

Given the simplicity and widespread use of the ShapeDesc struct, it has been moved to the same header file as Shape (regression/common/shape_utils.h).  Since the ShapeDesc encapsulates all properties required to define a Shape, the Shape::parseObj() factory method has also been overloaded to accept a ShapeDesc.

All examples and regressions using local definitions of ShapeDesc now use the common definition, and where Shape::parseObj() is used in the context of ShapeDesc instances, usage is replaced with the new Shape::parseObj(ShapeDesc const &).  Previous use of Shape::parseObj() was error prone and flawed in many cases as the full set of properties associated with a ShapeDesc were not always passed as arguments (particularly the left-handed option).

The "axis" argument to Shape::parseObj() was also removed, as all remaining usage expects the default value that preserves the original vertex positions.  Any required transformations are now expected to be applied externally and not during parsing.


c64kernal:
This is a very nice clean up, thank you @barfowl

PR #1095, opened 2019-04-24
Fixed single-crease patch creation near unisolated boundaries

barfowl:
This change adds newly required feature isolation for the single-crease patch that is now necessary in the absence of the previous boundary isolation for quads (removed in #1025).

When the single-crease patch option is used without the inf-sharp patch option, semi-sharp single-crease patches are intended to approximate interior inf-sharp edges.  When too many inf-sharp features (including boundaries) are present around a base face, despite the face being regular and easily and accurately represented with a single regular patch, isolation is required if any of the inf-sharp edges are interior edges to accommodate the single-crease patch approximation.


davidgyu:
Thanks, Barry!

PR #1096, opened 2019-04-25
Updated GL examples for common command line arguments

barfowl:
These changes update the remaining GL examples to make use of the common command line argument parsing and utilities recently added in #1087.  All now make use of common argument parsing and loading of user specified geometry.  Viewers that accept user-specified geometry were also updated to implement the -yup option consistently.

A couple of examples received some extra work:  glEvalLimit had additional UI options added to make it more consistent with glViewer, and glPtexViewer was updated to make use of the common ObjAnim class to support loading and interpolation of animated sequences of Obj files.

### DIFF
--- a/examples/glEvalLimit/glEvalLimit.cpp
+++ b/examples/glEvalLimit/glEvalLimit.cpp
@@ -80,6 +80,8 @@ GLFWmonitor* g_primary=0;
 #include <opensubdiv/far/error.h>
 
 #include "../../regression/common/far_utils.h"
+#include "../common/argOptions.h"
+#include "../common/viewerArgsUtils.h"
 #include "../common/stopwatch.h"
 #include "../common/simple_math.h"
 #include "../common/glControlMeshDisplay.h"
@@ -106,7 +108,8 @@ enum KernelType { kCPU = 0,
                   kGLXFB = 5,
                   kGLCompute = 6 };
 
-enum EndCap      { kEndCapBSplineBasis,
+enum EndCap      { kEndCapBilinearBasis,
+                   kEndCapBSplineBasis,
                    kEndCapGregoryBasis };
 
 enum HudCheckBox { kHUD_CB_DISPLAY_CONTROL_MESH_EDGES,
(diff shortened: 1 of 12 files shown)


examples/glEvalLimit/glEvalLimit.cpp
```diff
@@ -1397,12 +1408,28 @@ initHUD() {
     g_hud.AddPullDownButton(shading_pulldown, "FaceVarying", kFACEVARYING, g_drawMode==kFACEVARYING);
     g_hud.AddPullDownButton(shading_pulldown, "Mean Curvature", kMEAN_CURVATURE, g_drawMode==kMEAN_CURVATURE);
 
-    g_hud.AddCheckBox("Adaptive (`)", g_adaptive != 0, 10, 190, callbackCheckBox, kHUD_CB_ADAPTIVE, '`');
+    g_hud.AddCheckBox("Adaptive (`)", g_adaptive != 0, 10, 150, callbackCheckBox, kHUD_CB_ADAPTIVE, '`');
+
+    g_hud.AddCheckBox("Smooth Corner Patch (O)", g_smoothCornerPatch!=0,
+                      10, 170, callbackCheckBox, kHUD_CB_SMOOTH_CORNER_PATCH, 'o');
+//  g_hud.AddCheckBox("Single Crease Patch (S)", g_singleCreasePatch!=0,
+//                    10, 190, callbackCheckBox, kHUD_CB_SINGLE_CREASE_PATCH, 's');
```
c64kernal: Is this commented-out code intentional?

c64kernal:
Thanks so much @barfowl -- this is another really great clean up.  I left one small question.

examples/glEvalLimit/glEvalLimit.cpp
davidgyu: Commenting this out makes sense for glEvalLimit since single crease patches are currently supported for drawing only (not yet via the patch evaluation code paths).

The rest of the changes look good to me. Thanks, Barry!

PR #1099, opened 2019-05-02
Fixed bugs constructing patches with non-manifold features

barfowl:
These changes fix a bug that was not detected when construction of patches with non-manifold vertices was modified to use irregular patches (#1013), along with another bug specific to Loop patch construction along non-manifold edges.  Additional comments were added and revisions made to others that were out of date.

A second commit includes new regression shapes to provide better coverage of non-manifold topology in future.


davidgyu:
Thanks, Barry! This looks great. Thank you for the additional test cases too!

PR #1100, opened 2019-05-06
Added new Catmark test shapes to regression/shapes

barfowl:
This commit includes several new test shapes for the Catmark scheme.  Two provide additional coverage of non-manifold features and correspond closely to shapes recently added for Loop.  Two others consolidate crease and corner features on cubes and similarly correspond to test shapes present for Loop.

These new collections of shapes make several existing shapes redundant and can be used to reduce the cumbersome shape lists in the example programs without reducing test coverage.


davidgyu:
Thanks, Barry!

PR #1102, opened 2019-05-08
Additions and revisions to regression/shapes for consistency

barfowl:
These changes include the addition of new shapes for the Catmark scheme -- some to match the conventions recently adopted for the Loop shapes, and others to provide more coverage of cases specific to Catmark.  Minor changes to a couple of recently added Loop shapes were also made to more closely match the corresponding Catmark cases.


davidgyu:
Thanks!

PR #1103, opened 2019-05-08
Transformed old regression/shapes into consistent orientation

barfowl:
Several old test shapes were added without any consideration of the viewing and testing applications that most make use of them.  Shapes are generally oriented in the XZ plane with the features they are intended to exhibit facing the default viewpoint along the Y axis.  Some shapes are actually oriented in the complete opposite direction.

These changes transform shapes that deviated most into this common orientation for consistency -- most requiring a simple rotation of 180 or 90 degrees about some axis.  Aside from consistency in the example viewers, this will also make the batch imaging tests we hope to apply more effective.

Baseline data for the old hbr_regression test was also updated for the shapes that were modified (here and previously) to keep it running clean.


davidgyu:
Nice! Thank you, Barry!

PR #1104, opened 2019-05-14
Fixed patch construction bug with sharp corner patches

barfowl:
This change fixes an uncommon bug that occurs when building a patch from a face when both ends of an interior edge are infinitely sharp.  In this case, the stencil size of the interior face-points was not correctly estimated and triggers an assertion.

This can show up when vertices are implicitly made inf-sharp, e.g. when vertices are non-manifold or when boundary vertices in face-varying space are sharpened according to the interpolation choice.


davidgyu:
Looks good to me. Thanks, Barry!

PR #1105, opened 2019-05-14
Added face-varying UVs to several regression shapes

barfowl:
These changes add UVs to several sets of shapes in regression/shapes to provide more coverage of face-varying patch construction.


davidgyu:
Thanks!

PR #1106, opened 2019-05-15
Fixed indexing errors in old regression/shapes

barfowl:
Several test shapes in regression/shapes have index errors that can cause some subtle problems when testing or limit their use elsewhere.  In a couple of cases, the number of texture coordinates provided is one short of what is specified in the face definitions.  In most cases though, the face definitions refer to texture or vertex normals that do not exist at all.

These changes clean up these files -- adding a single missing entry when necessary or stripping texture or normal indices from face definitions when none are present or they are completely uncorrelated.

The question of where these errors should be detected in future, e.g.  during parsing of the Obj format, subsequent validation of the Shape that results, etc., is left for another time.


davidgyu:
Thank you for taking care of all of these!

PR #1107, opened 2019-05-16
Minor improvements to examples/glFVarViewer

barfowl:
Changes here include several improvements to the glFVarViewer for inspecting face-varying interpolation and patches via UVs:
* added check-boxes for Control Edges and Control Vertices (common to other examples)
* added check-box for UV Backface Culling (defaults to Off so all UVs seen by default)
* relabelled the Boundaries pull-down to Linear Interpolation to reflect the enum it affects

Some of the source was additionally modified to avoid use of "boundary" to avoid confusion with the VtxBoundaryInterpolation option that is not affected.


davidgyu:
Excellent! Thanks, Barry!

PR #1108, opened 2019-05-16
Fixed refinement bug for non-manifold edges with many incident faces

barfowl:
This change fixes a bug present since 3.0 where a large number of faces share a common non-manifold edge.  The edge-vertex introduced on this edge in the first level of refinement can have higher valence than the maximum currently determined for the base level.

The "max valence" member is mainly used internally and externally for sizing memory so overflows can occur if it is under-estimated.  For quad refinement the under-estimation will be at most one, but for triangles the new max valence can be double the original (since two edges are introduced for every incident face).

Two new regression shapes were added to stress this condition in future.  Per-vertex UVs were included in these shapes as building a face-varying StencilTable under these conditions is the use case that first drew attention to the overflow.


jtran56:
Filed as internal issue #OSD-289.

davidgyu:
Good catch! And thanks again for the addition test cases!

PR #1109, opened 2019-05-22
Fixed regression with legacy smooth corner patch approximation

barfowl:
This change fixes a regression that was introduced into the application of the legacy patch construction option that approximates a smooth corner patch with a regular sharp patch.  The condition of application had been rewritten and mistakenly considered all corners, rather than only those that had been determined as smooth.  So some sharp corners were being mistakenly re-assigned as smooth.

This tended to occur more frequently in face-varying patches, where the combination of inf-sharp and irregular features on boundaries is more prevalent along the UV seams.


davidgyu:
Nice catch. Thanks, Barry!

PR #1110, opened 2019-05-23
Fixed semi-sharp face-varying boundaries when creating inf-sharp patches

barfowl:
This change fixes an obscure bug present since 3.1 when face-varying patches and the inf-sharp patch option were introduced.  Semi-sharp features on the seams of face-varying patches were mistakenly tagged as inf-sharp, which only took effect when the inf-sharp patch option was enabled.


jilliene:
Filed as internal issue #OSD-290

PR #1112, opened 2019-05-31
Added tetrahedron shape use in documentation images and logo

barfowl:
This addition includes Catmark and Loop versions of the tetrahedral shape used in some of the documentation pages and the logo.  Both shapes include UVs -- with three disjoint UV regions meeting at the valence-3 corners, its a good example to illustrate the difference between the CORNERS_ONLY and CORNERS_PLUS face-varying interpolation modes.


davidgyu:
Excellent! Thanks, Barry!

PR #1114, opened 2019-06-05
Updated shape lists for example viewers

barfowl:
These changes update the list of shapes shown in the three example viewers: glViewer, dxViewer and mtlViewer.  The new lists include several recently added shapes, remove redundant shapes from the Catmark set and reorder the lists for consistency between the Catmark and Loop sets.


davidgyu:
Great! Thanks!

PR #1115, opened 2019-06-05
Updated shape list for GL face-varying viewer

barfowl:
This change updates the list of shapes used to view face-varying UVs with examples/glFVarViewer.  Several shapes that were recently merged were included while some shapes that have since been made redundant were removed.


jilliene:
Filed as internal issue #OSD-291

PR #1116, opened 2019-06-08
Improved regular patch approximations of irregular features for Loop

barfowl:
This change improves the shape of irregular patches for Loop when specifying the use of regular (Box-spline) patches rather than the Gregory triangles that enforce tangent continuity.


davidgyu:
This is great! Thanks, Barry!

PR #1117, opened 2019-06-08
Minor revisions to a few precision related methods of Far classes

barfowl:
These changes modify a couple of public methods added to Far to support variable precision based on feedback.  Methods to query the precision of Far::PatchTable's internal StencilTables were renamed, along with the precision related members of Far::PatchTableFactory::Options.


davidgyu:
Thanks, Barry!

PR #1123, opened 2019-06-14
Fixed Loop patch creation bug with val-3 vertex at third corner

barfowl:
This change fixes a bug creating triangular patches for Loop subdivision when a valence-3 interior vertex occurs specifically at the third and last corner of the face.

A second commit modifies the test shapes that focus on extra-ordinary vertices to rotate the first few faces -- putting the extra-ordinary vertex at each corner -- to better cover these kinds of cases in future.


davidgyu:
Excellent! Thanks, Barry! Thank you for the updated test cases as well!

PR #1124, opened 2019-06-14
Added simple tetrahedron to set of Loop test shapes

barfowl:
A tetrahedron is added here to the set of test shapes for Loop.  An open tet (one face missing) and a tet with holes (the toroidal_tet) are already present, but a simple tetrahedron has been absent.

A second commit includes the new shape in the viewer examples for the main back-ends, along with the glFVarViewer for inspecting its assigned UVs.


davidgyu:
Thanks!

PR #1125, opened 2019-06-17
Improvements to far_perf output and options

barfowl:
These changes build on #1119 in providing more flexibility with regression/perf.  Changes include:

- fixing memory leaks and asserting or reporting bad input data
- re-organizing the CSV output and renaming the option to -cvs (from -spreadsheet)
- addition of the -nostencils and -nopatches options to reduce and focus execution and output
- addition of the -total flag to provide only a summary of each test

Several of the growing set of options are also factored into two groups -- those affecting execution of the tests and those affecting the output generated.


davidgyu:
Excellent! Thanks, Barry!

PR #1126, opened 2019-06-18
Minor improvements to examples for Loop

davidgyu:
These improve some of the example shader code related to
adaptive refinement and tessellation of Loop meshes.

The main change is to update shader configuration so that it
depends only on the resulting patch type. Previously, there
were cases where a "LOOP" preprocessor symbol was passed
from the application to the shader code, but this is error prone
and confusing.

Also, fixed a few minor bugs discovered during testing.


c64kernal:
Nice cleanups, thanks @davidgyu

PR #1127, opened 2019-06-18
Examples options and shapes

davidgyu:
This makes the option default settings more consistent for the various examples
and updates the shape lists for several of the examples match the current
shape list in the glViewer, etc.


c64kernal:
Looks good, thanks @davidgyu

PR #1128, opened 2019-06-18
Update glFVarViewer option defaults

davidgyu:
Enabled smooth corner, inf sharp and single crease patch
options by default.


c64kernal:
Thanks @davidgyu

PR #1129, opened 2019-06-18
Fixed cracks in face-varying Loop patches for lineaar-boundaries option

barfowl:
This change addresses discontinuities that appear between face-varying patches for Loop when using the LINEAR_BOUNDARIES option.  The change works around a deeper issue in the way that face-varying topology is tagged for this option -- the boundary vertices are tagged as corners for ease of refinement and interpolation, but need to be re-assessed when considered as part of a patch.


davidgyu:
Good catch! Thanks, Barry!

PR #1136, opened 2019-06-21
Improved examples/glPaintTest with slider to modify brush size

barfowl:
The fixed brush size of 500 pixels in examples/glPaintTest wasn't very effective in demonstrating its effects.  These changes set the default brush size to a more reasonable 100 pixels and add a slider to modify the brush size interactively.


davidgyu:
Nice! Thanks!

PR #1141, opened 2019-06-24
Added and revised figures in the introductory documentation

barfowl:
Added a new pair of figures in the introductory "Subdivision Surfaces" section to illustrate the concept of a basis function.  Replaced the stand-in image that was used for the poses of an animated mesh with three unique poses.


davidgyu:
These are great! Thanks, Barry!

PR #1147, opened 2019-06-27
Improved Far patch parameterization documentation for non-quads

barfowl:
These changes improve the section on Patch Parameterization in the Far Overview to illustrate the quadrangulation of non-quad faces and the offset indices that result -- requiring the use of Far::PtexIndices.


davidgyu:
Thank you, Barry!

Issue #1157, opened 2019-08-14
CMake configuration error when using vcpkg on Windows


zomgrolf:
Hi, my apologies for bothering you with something so trivial.

I'm getting a CMake configuration error trying to build OpenSubdiv on Windows.

The reason is, I'm trying to use Microsoft's vcpkg to deal with the dependencies: the idea is that I can easily build the necessary dependencies using vcpkg, then I can simply point CMake to the right toolchain file (provided by vcpkg), so that CMake scripts for OpenSubdiv can easily find what they need, as per [vcpkg integration docs](https://github.com/microsoft/vcpkg/blob/master/docs/users/integration.md)). 

Works amazingly well, apart from one little gotcha: the main CMakeLists.txt in OpenSubdiv defines a macro `_add_executable` [macro](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/9be5a298270a42d89d7925c8cdd9e335869ec6ce/CMakeLists.txt#L677), which clashes with custom scripts used by vcpkg: apparently, when overriding a function implementation in CMake, the original version gets renamed by adding the underscore prefix, and it happens that vcpkg overrides `add_executable` in order to do its thing -- that ends up clashing with the custom macro defined by OpenSubdiv. I tried renaming the macro to `osd_add_executable` and that fixed the issue for me (I also had to edit CMakeLists.txt in [stringify](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/9be5a298270a42d89d7925c8cdd9e335869ec6ce/opensubdiv/tools/stringify/CMakeLists.txt#L32)).

I could create a vcpkg port of OpenSubdiv and patch it there, but it's such a tiny change and it's probably not a good idea to name your macro `_add_executable` in the first place, so I thought I should open a new issue.

jilliene:
Filed as internal issue #OSD-301

c64kernal:
Thank you so much for raising this @zomgrolf  -- we should absolutely fix this instance and likely scrub a lot of the other macro names.

davidgyu:
Thanks again! We believe these macros now have all been fixed to have an "osd" prefix.

PR #1159, opened 2019-08-20
Prefix OpenSubdiv CMake macros

c64kernal:
This fixes issue #1157 by adding "osd" as a prefix to public macros.


davidgyu:
Thanks for the fix, George!

PR #1167, opened 2019-12-14
Moved common command line parsing to regression/common

barfowl:
These changes move the ArgOptions class for parsing common command line arguments from examples/common to regression/common.  This allows it to be used in regression tests and tutorials without creating the link dependencies that examples/common introduces (e.g. osdGPU, Ptex, etc.).


davidgyu:
Thanks, Barry!

PR #1168, opened 2019-12-15
Adding Azure Pipelines support for OSD

c64kernal:
This commit adds initial Azure Pipelines CI support for OpenSubdiv.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,50 @@
+trigger:
+- master
+- dev
+
+jobs:
+- job: Linux
+  pool:
+    vmImage: 'Ubuntu-16.04'
+  steps:
+  - script: |
+      sudo apt-get -qq update;
+      sudo pip install --upgrade pip
+      sudo apt-get install python-setuptools;
+      sudo apt-get install libglew-dev libxrandr-dev libxcursor-dev libxinerama-dev libxi-dev;
+      sudo python build_scripts/build_osd.py --tbb --omp --build $HOME/OSDgen/build --src $HOME/OSDgen/src $HOME/OSDinst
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,50 @@
...
+    inputs:
+      artifactName: 'osd-macOS'
+      targetPath: "  /Users/runner/OSDinst"
+
+- job: Windows
+  pool:
+    vmImage: 'VS2017-Win2016'
+  steps:
+  - script: |
+      call C:\"Program Files (x86)"\"Microsoft Visual Studio"\2017\Enterprise\VC\Auxiliary\Build\vcvars64.bat
```
thomthom: Might be good to test build with static and dynamic runtimes?

davidgyu: Yes, good suggestion! We'll take a look at that!

davidgyu:
This is great! Thanks, George!

azure-pipelines.yml
c64kernal: +1! As David said, we can start from here and build up with lots more configurations too...

thomthom: Sounds good to me. (I just recalled having issues with static CRT in the past.)

Curiously, this project is using CMake, but you are calling independent build scripts here. Any particular reason for that?

c64kernal: The build script just downloads the necessary dependencies and invokes cmake -- so it's all still cmake, but the build script makes the dependency management easier and keeps the .yml files small :)

thomthom: Gotcha, I didn't look too close at the script before commenting.

PR #1170, opened 2020-01-14
Added selective interpolation for VtxBoundaryInterpolation case "none"

barfowl:
The general behaviour for Sdc::Options::VtxBoundaryInterpolation case "none" is to exclude the limit surface for all boundary faces (achieved by internally tagging them as holes).  But apparently the expected behaviour of this feature in RenderMan allows for the selective interpolation of boundary faces by making their neighbouring boundary edges infinitely sharp.  This is not widely documented (if at all) but is a relatively easy addition to maintain compatibility with RenderMan.

Changes here include:

    - recognition of user-assigned boundary edge sharpness in the tagging of boundary faces
    - new test shapes for Catmark and Loop to illustrate the added behavior
    - changes to test shapes for other boundary interpolation options for consistency
    - revisions to the documentation/Doxygen to reflect the added behavior

All of the test shapes for the VtxBoundaryInterpolation cases were updated to more clearly reflect illustrations in the user documentation.  The simple 16-point grid of quads used for each case in the figures is now included in the test shapes for Catmark, with a similar shape included for all of the Loop cases.


davidgyu:
Great! Thanks, Barry!

PR #1171, opened 2020-01-14
Updated examples with new boundary interpolation test shapes

barfowl:
As a follow-up to #1170, this change adds the new test shapes created for the boundary interpolation case "none" to all of the interactive examples.

In order to keep the shapes lists from growing endlessly, some shapes that have recently been made redundant by improvements to others were removed to partially offset the additions.


davidgyu:
Thanks!

PR #1173, opened 2020-01-15
Updated cmake/Find* modules for VS2019

davidgyu:
For Visual Studio 2019 the target architecture is specified
separately from the generator, and this updates the cmake
modules to be able to locate libraries for the correct target.

Also, makes use of CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION
to be able to use the correct Windows Kit when multiple
tool sets and development kits are available.


c64kernal:
Thanks for the fix, looks good!

PR #1174, opened 2020-01-15
Fixed Ptex link dependencies

davidgyu:
Removed the link dependency from libosd to
the Ptex library since Ptex is used only by the
examples. Updated examples/common to separate
the common Ptex sources from the other common
sources so that they are used only by the Ptex
examples.


jilliene:
Filed as internal issue #OSD-315

c64kernal:
This is very nice, thanks @davidgyu!

PR #1175, opened 2020-01-22
Fixed memory leak in far/tutorial_5_2 with optional .obj file

barfowl:
The main change here fixes a memory leak when an optional .obj file was specified.  Given this tutorial was written with parameters to illustrate time versus space trade-offs, a memory leak here is potentially more of an issue than in others.

Other minor changes were applied while updating the source: command line parsing was updated to use tools in regression/common, and line-wrapping and indentation were updated to conform to standards.


davidgyu:
Thanks, Barry. Good catch!

PR #1176, opened 2020-01-25
Added new tutorial for Far::LimitStencilTable

barfowl:
Example usage of Far::LimitStencilTable was added in tutorials/far/tutorial_5_3.  Stencils for limit points of a crude tessellation of a shape are computed and applied to a series of translated poses of that shape.

Since LimitStancilTables require understanding of patches and PatchTables, it was added to the 5.x tutorials that cover PatchTables rather than the 4.x tutorials that cover StencilTables.


davidgyu:
This is a great addition to the tutorials. Thanks, Barry!

PR #1177, opened 2020-02-11
Fixed initialization oversight in far/tutorial_5_1

barfowl:
Prior to the release of 3.4, the newly added Far::PatchTableFactory method to derive adaptive refinement options from patch table options was illustrated in far/tutorial_5_1.  But the desired refinement level was not being used to initialize the patch options.  As a result, the default value of 10 was being used to initialize adaptive refinement options.

The changes here initialize the patch options with the desired level and include a comment to warn about this kind of oversight in future.


jilliene:
Filed as internal issue #OSD-317

davidgyu:
Thanks!

PR #1182, opened 2020-03-03
Fixed strict-aliasing warnings

davidgyu:
Minor update to the implementation of a few internal tagging
methods in Vtr to avoid violating strict-aliasing rules.
Specifically, switched to using std::memcpy instead of
reinterpret_cast to implement type punning.


c64kernal:
Looks good, thanks @davidgyu

PR #1192, opened 2020-04-01
Fixed static library linking on Windows

davidgyu:
Changed the glLoader library to an object library
instead of a static library to improve generating
opensubdiv static libraries.


c64kernal:
Thanks @davidgyu

PR #1205, opened 2020-10-27
Python 3 updates to build script

c64kernal:
These updates should allow the build_osd.py script to work with Python 3.x and is based on changes made to the similar script in the USD distribution.


davidgyu:
Looks good. Thanks, George!

PR #1206, opened 2020-10-28
Adjustments to CMake config for use as a sub-project.

thomthom:
I've been using OpenSubdiv in my project as a git submodule. Right now I'm in the process of converting my project to use CMake, so I absorbed OpenSubdiv into the build by using `add_subdirectory('ThirdParty/OpenSubdiv')`

While doing so I found that my build output for the whole project ended up under `ThirdParty/OpenSubdiv/bin` and `ThirdParty/OpenSubdiv/lib`.

After poking around I found that OpenSubdiv set two global properties that affect not only it's own dir and descendants, but also parent projects and directories - everything in the build.

This PR guard against setting these properties when the project is consumed as a sub-project.

While I was at it I also fixed a reference from `CMAKE_PROJECT_NAME` (which picked up the root project name) to `PROJECT_NAME` (Name of the last `project` call)

---

For reference, where is a fragment of CMake configuration output before my changes:
(I dump a few variables before I invoke `add_subdirectory` and then again after.

```
[cmake] -- ----------
[cmake] -- CMAKE_INSTALL_PREFIX: C:/Program Files (x86)/SUbD
[cmake] -- CMAKE_BINDIR_BASE:
[cmake] -- LIBRARY_OUTPUT_PATH_ROOT:
[cmake] -- EXECUTABLE_OUTPUT_PATH:
[cmake] -- LIBRARY_OUTPUT_PATH:
[cmake] -- ----------
[cmake] -- Compiling SUbD version v3_4_3
[cmake] -- Using cmake version 3.17.2
[cmake] -- Using static MSVC CRT
[cmake] -- ----------
[cmake] -- CMAKE_INSTALL_PREFIX: C:/Program Files (x86)/SUbD
[cmake] -- CMAKE_BINDIR_BASE:
[cmake] -- LIBRARY_OUTPUT_PATH_ROOT:
[cmake] -- EXECUTABLE_OUTPUT_PATH: C:/Users/Thomas/Source/SUbD/build/ThirdParty/OpenSubdiv/source/OpenSubdiv/bin
[cmake] -- LIBRARY_OUTPUT_PATH: C:/Users/Thomas/Source/SUbD/build/ThirdParty/OpenSubdiv/source/OpenSubdiv/lib
[cmake] -- ----------
```

After my changes:

```
[cmake] -- ----------
[cmake] -- CMAKE_INSTALL_PREFIX: C:/Program Files (x86)/SUbD
[cmake] -- CMAKE_BINDIR_BASE: 
[cmake] -- LIBRARY_OUTPUT_PATH_ROOT: 
[cmake] -- EXECUTABLE_OUTPUT_PATH: 
[cmake] -- LIBRARY_OUTPUT_PATH: 
[cmake] -- ----------
[cmake] -- Compiling OpenSubdiv version v3_4_3
[cmake] -- Using cmake version 3.17.2
[cmake] -- Using static MSVC CRT
[cmake] -- ----------
[cmake] -- CMAKE_INSTALL_PREFIX: C:/Program Files (x86)/SUbD
[cmake] -- CMAKE_BINDIR_BASE: 
[cmake] -- LIBRARY_OUTPUT_PATH_ROOT: 
[cmake] -- EXECUTABLE_OUTPUT_PATH: 
[cmake] -- LIBRARY_OUTPUT_PATH: 
[cmake] -- ----------
```

And the build output is located where I would expect it to be.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,10 +53,17 @@ endif()
 
 #-------------------------------------------------------------------------------
 
-message(STATUS "Compiling ${CMAKE_PROJECT_NAME} version ${OpenSubdiv_VERSION}")
+message(STATUS "Compiling ${PROJECT_NAME} version ${OpenSubdiv_VERSION}")
 message(STATUS "Using cmake version ${CMAKE_VERSION}")
 
 #-------------------------------------------------------------------------------
+# Determine if the project is built as a subproject (using add_subdirectory)
+# or if it is the master project.
+set(MASTER_PROJECT OFF)
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    set(MASTER_PROJECT ON)
+endif()
+
 # Specify the default install path
 if (NOT DEFINED CMAKE_INSTALL_PREFIX)
     SET( CMAKE_INSTALL_PREFIX "${PROJECT_BINARY_DIR}/" )
@@ -97,19 +104,21 @@ if(LIBRARY_OUTPUT_PATH_ROOT)
     SET( CMAKE_INSTALL_PREFIX "${LIBRARY_OUTPUT_PATH_ROOT}/" )
 endif()
 
-# Set the directory where the executables will be stored.
-set(EXECUTABLE_OUTPUT_PATH
-    "${PROJECT_BINARY_DIR}/bin"
-    CACHE PATH
-    "Directory where executables will be stored"
-)
+if (MASTER_PROJECT)
+    # Set the directory where the executables will be stored.
+    set(EXECUTABLE_OUTPUT_PATH
+        "${PROJECT_BINARY_DIR}/bin"
+        CACHE PATH
+        "Directory where executables will be stored"
+    )
 
-# Set the directory where the libraries will be stored.
-set(LIBRARY_OUTPUT_PATH
-    "${PROJECT_BINARY_DIR}/lib"
-    CACHE PATH
-    "Directory where all libraries will be stored"
-)
+    # Set the directory where the libraries will be stored.
+    set(LIBRARY_OUTPUT_PATH
+        "${PROJECT_BINARY_DIR}/lib"
+        CACHE PATH
+        "Directory where all libraries will be stored"
+    )
+endif()
 
 # Specify the list of directories to search for cmake modules.
 set(CMAKE_MODULE_PATH
@@ -652,7 +661,7 @@ if (WIN32)
             "environment variable."
         )
     endif()
-    
+
     # Link examples & regressions statically against Osd for
     # Windows until all the kinks can be worked out.
     if( OSD_GPU )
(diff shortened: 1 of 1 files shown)


thomthom:
Looks like the CI build is failing unrelated to this PR. Syntax error in a script by the looks of it.

image

thomthom:
hmmm... wait. this appear to be fixed in `dev`. I guess I should rebase and make my PR against that branch.

jilliene:
Filed as internal issue #OSD-339

davidgyu:
Thanks @thomthom This looks very useful and we'd like to consider including this.

Do you know of other examples or documentation of the method for subproject detection in addition to the repo link you mentioned in the comment?

thomthom:
> Do you know of other examples or documentation of the method for subproject detection in addition to the repo link you mentioned in the comment?

I've been getting familiar with CMake over the last 1-2 years (I think...) and my main source of learning resource has been Craig Scott's [Professional CMake: A Practical Guide](https://crascit.com/professional-cmake/). It's been a great guide to weed out what is old CMake practices from newer more practical workflows.

The pattern I suggest in this PR is similar to what he describes in the book. But I've also used the `{fmt}`' project as reference for a real world complete project: https://github.com/fmtlib/fmt/blob/8c8f74a870bbf2d9de726bc347d05831cdc01da7/CMakeLists.txt#L12

Another key find I find useful with CMake project is when they use target properties. Such that you only need to add the name of the target and include paths etc will automatically carry over.

At the moment I need to add the location of the OSD headers manually for each target that links with OSD, which isn't needed if the target had been setup with `target_include_directories`. However, this might not be compatible with the current minimum CMake version of this project. (Maybe there is a way to conditionally allow this for users of newer CMake.) Since CMake~ 3.15 there's been a great host of useful new features. That might be a version to recent as a requirement for a library like this, but maybe it could be considered to be something higher than the current 2.8.6. I couldn't find any notes on why the current minimum CMake version is what is it is now, so it's hard for me to make the best suggestions.

I was going to make another issue with some suggestions once I'd gotten things working as they are. If you could shed some light on your minimum CMake requirements are it's make it easier to work out a proposal.

davidgyu:
Great notes! Thank you!

Agree that it could be good to be able to use facilities available in more recent versions of CMake. In practice we've been using CMake 3.x for most of our local testing and deployment.

davidgyu:
Could you update this to remove the outbound URL at line 62? Otherwise looks great and we are ready to merge.

We've discussed updating our minimum CMake requirements, and we're going to defer that for now, but we are interested in cleaning up and improving the build and welcoming suggestions. Thanks again!

thomthom:
> Could you update this to remove the outbound URL at line 62? Otherwise looks great and we are ready to merge.

Sure thing!

> We've discussed updating our minimum CMake requirements, and we're going to defer that for now, but we are interested in cleaning up and improving the build and welcoming suggestions. Thanks again!

I'll look closer at what improvements can be made to make this project consumed as a sub-module within the context of CMake 2.8.6.

davidgyu:
Thanks!

PR #1207, opened 2020-12-02
Fixed bug in double precision template for Far::StencilTableFactory

barfowl:
This change fixes a bug in the double precision template for Far::StencilTableFactory, i.e. Far::StencilTableFactoryReal<double>.

The Create() method uses the Far::PrimvarRefiner to assemble the stencils -- another class that is parameterized by precision.  But the template version of PrimvarRefiner was not declared with the appropriate precision parameter here.  Instead the backward-compatible single-precision version was used, which compromises the accuracy of refined vertices.


jilliene:
Filed as internal issue #OSD-340

PR #1208, opened 2020-12-02
Fixed Far::TopologyRefiner::GetMaxLevel() after call to Unrefine()

barfowl:
This change fixes the reassignment of the _maxLevel member of Far::TopologyRefiner as part of the call to Unrefine().

The _maxLevel member was being left unmodified rather than being reset to 0, causing GetMaxLevel() to return an incorrect result.  Meanwhile the GetNumLevels() method returned the correct result before and after and so provides a simple workaround in place of GetMaxLevel().


jilliene:
Filed as internal issue #OSD-341

Issue #1201, opened 2020-10-02
build failure under CMake daily build


meshula:
In order to build with Xcode 12.2, the upcoming CMake 3.19 is required to generate correct Xcode project files.

CMake now treats a condition as an error that was previously not an error. Specifically, when generating frameworks for MacOS, and when a shared library is requested, both frameworks are created to the same path, due to:

```
        set_target_properties(
            osd_static_framework
            PROPERTIES
            FRAMEWORK true
            INSTALL_NAME_DIR "@rpath/OpenSubdiv.framework/OpenSubdiv"
            INSTALL_RPATH "@executable_path/Frameworks;@loader_path/Frameworks"
            OUTPUT_NAME OpenSubdiv
            CLEAN_DIRECT_OUTPUT true
        )

and

            set_target_properties(
                osd_dynamic_framework
                PROPERTIES
                RPATH true
                FRAMEWORK true
                INSTALL_NAME_DIR "@rpath/OpenSubdiv.framework/OpenSubdiv"
                INSTALL_RPATH "@executable_path/Frameworks;@loader_path/Frameworks"
                OUTPUT_NAME OpenSubdiv
                CLEAN_DIRECT_OUTPUT true
            )

```

CMake now treats it as an error that two targets have the same output. An easy fix is to change the OUTPUT_NAME of the shared library to be unique, e.g. OpenSubdiv_shared, or to enforce that either a shared or static library may be built but not both. It would seem least intrusive and bothersome to simply adopt a new name.

jilliene:
Filed as internal issue #OSD-331

davidgyu:
Fixed by #1214

PR #1214, opened 2021-02-03
Fixed macOS static framework installation

### LINKED ISSUE #1194: Intermittent build failures on MacOS during framework creation
Building OSD on MacOS occasionally fails with errors like the following:

> cd /Users/usdonmac/Projects/usd/sunya/osd_test/build/OpenSubdiv-3_3_3/opensubdiv && /Applications/CMake.app/Contents/bin/cmake -E cmake_symlink_library ../lib/libosdGPU.3.\
3.3.dylib ../lib/libosdGPU.3.3.3.dylib ../lib/libosdGPU.dylib
cmake -E touch_nocreate: failed to update "lib/OpenSubdiv.framework/OpenSubdiv".
make[2]: *** [lib/OpenSubdiv.framework/OpenSubdiv] Error 1
make[1]: *** [opensubdiv/CMakeFiles/osd_static_framework.dir/all] Error 2

Looking through the full log (attached), it appears that the build is trying to write both the static and dynamic libraries to the same file in the framework:

> [100%] Linking CXX shared library ../lib/libosdGPU.dylib
[100%] Linking CXX static library ../lib/OpenSubdiv.framework/OpenSubdiv
[100%] Linking CXX static library ../lib/libosdGPU.a
[100%] Linking CXX shared library ../lib/OpenSubdiv.framework/OpenSubdiv

My guess is that the build may be trying to write to the same file concurrently, leading to the error above.

I've reproduced this on 3.3.0 and the 3.4.1+ release candidates when building with make using -j 12 but not with -j1.

davidgyu:
Assigned the postfix "_static" to the static framework
and also fixed header installation. This makes the static
framework usable and also fixes build errors when using
recent versions of CMake. Also fixes build errors that
can occur when building with multiple threads.

Fixes #1194 #1201

### DIFF
--- a/opensubdiv/CMakeLists.txt
+++ b/opensubdiv/CMakeLists.txt
@@ -287,7 +287,7 @@ if (NOT NO_LIB)
             FRAMEWORK true
             INSTALL_NAME_DIR "@rpath/OpenSubdiv.framework/OpenSubdiv"
             INSTALL_RPATH "@executable_path/Frameworks;@loader_path/Frameworks"
-            OUTPUT_NAME OpenSubdiv
+            OUTPUT_NAME OpenSubdiv_static
             CLEAN_DIRECT_OUTPUT true
         )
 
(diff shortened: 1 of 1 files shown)


jilliene:
Filed as internal issue #OSD-349

c64kernal:
Looks good to me, thanks @davidgyu

davidgyu:
Testing notes -- verified on the following systems using multithreaded builds, both as a standalone OpenSubdiv build and as a part of a USD build after removing the single threaded build workaround from build_usd.py:
  macOS 10.14.6 Mojave, Xcode 11.3.1, CMake 3.16.5
  macOS 10.15.7 Catalina, Xcode 12.4, CMake 3.19.3
  macOS 11.2 Big Sur, Xcode 12.4, CMake 3.19.3

PR #1216, opened 2021-02-03
Merged GL Api Loader fixes from USD

davidgyu:
Merged recent bug fixes from the USD implementation of the
GL Api Loader. This fixes the OpenGL system framework path on
macOS and also improves GL version and GL extension processing.


c64kernal:
Looks good, thanks @davidgyu

PR #1217, opened 2021-02-03
Removed obsolete references to hbr from farViewer

davidgyu:


c64kernal:
Very nice find, thanks @davidgyu

PR #1220, opened 2021-02-04
Additional fix/cleanup for FindDocutils.cmake

davidgyu:
Always call FindPackageHandleStandardArgs() and allow it to
set DOCUTILS_FOUND appropriately. By marking DOCUTILS_VERSION
as required we'll also get a useful error message if the version
string from RST2HTML_EXECUTABLE could not be parsed correctly.


c64kernal:
Nice clean up, @davidgyu

PR #1227, opened 2021-04-15
Added methods to access vtags _xordinary and _corner in topologyLevel.h

roquo:


jilliene:
Filed as internal issue #OSD-353

davidgyu:
Hi! Would you please submit a signed Contributor License Agreement as described here: [Contributing to OpenSubdiv](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/master/documentation/contributing.rst#contributor-license-agreement) Thank you!

roquo:
I have sent the CLA to opensubdiv-cla@pixar.com. Anything else to do?

barfowl:
Glad to hear you're still with us. After almost a year and a half of no response to the CLA we were considering closing this request.

These suggestions were discussed offline sometime last year and it was agreed that it's not appropriate to expose bits of the internal VTags directly.  They were defined for very specific internal purposes and are not all suitable for more general public queries.

I'll follow up shortly with details on how to redefine and document the two methods proposed here. But, since the original posting, I did prepare alternatives for these methods along with a few others that might be worthwhile.

So if you want to leave this up to us, we will get a new pull request opened and merged soon (within next week or so) that includes a set of additions that should meet both your needs and our standards and conventions.

More details to follow...

barfowl:
Given only two inline functions with one-line comments, I'll put the suggestions and their rationale here rather than per-line comments on the code...

**IsVertexCorner():**  The returned _corner tag is set when the vertex has only one incident face, is manifold, and is infinitely sharp. In contrast, the existing method IsVertexBoundary() reports if a vertex lies on a boundary regardless of whether it is smooth or sharp, manifold or non-manifold. We prefer these behave more consistently, i.e. the new test for corner is more similar to the existing test for a boundary.

So rather than checking the _corner tag, IsVertexCorner() should simply test the number of incident faces and return true when only one:
```
/// \brief Return if the vertex is a corner (one incident face)
bool IsVertexCorner(Index v) const { return (_level->getNumVertexFaces(v) == 1); }
```
It is up to the caller to further test if a corner vertex is smooth or sharp, manifold or not. etc.

**IsVertexExtraOrdinary():**  The situation here is a little more complicated as the returned_xordinary tag is not always defined.  It is only defined when the vertex is manifold, so at the very least, we need to clearly document that (as "not extra-ordinary" does not always mean "is regular")

We also prefer not to introduce the new term "ExtraOrdinary" into the naming of the public API as we already use the complementary term "Regular" in many other cases.

So that leads to IsVertexRegular() defined and documented as:
```
/// \brief Return if the valence of a manifold vertex is regular:
bool IsVertexRegular(Index v) const { return !_level->getVertexTag(v)._xordinary; }
```
It is up to the caller to ensure the vertex is manifold before making any assessment of regular versus extra-ordinary.

roquo:
@barfowl thanks for your reply. I forgot about this pull request because I did not use any new version of OpenSubdiv. As I wanted to use the latest version now I saw that I could not compile anymore because of the missing functions. Sorry about that!

It would be really great if you could add these methods to the library and it is totally ok if you are doing this. I then would close this pull request when the methods are available. Thanks!

barfowl:
This pull request has now been addressed by #1255.

A minor change from the plan noted here was that the name of the IsVertexRegular() method was changed to IsVertexValenceRegular() to emphasize that it refers to the valence only and not the neighborhood of the vertex. This was further emphasized in the accompanying Doxygen comments.

Thanks again for the suggestions.

roquo:
@barfowl thanks for all your effort!
Is there any plan for when the next release is going to happen? I would really appreciate it if you could give me some information on that. Then our development team can plan for our software release if we have to ship with the dev(main) branch or if we can use a released version. Thanks

davidgyu:
@roquo  We're preparing a new release now which will include these fixes. We're expecting to tag a release candidate early next week.

PR #1234, opened 2021-07-28
Support oneTBB 2021

### LINKED ISSUE #1211: Adding support for oneTBB

Logarithmus:
@jilliene

Closes #1211

I'm not sure my solution is right, I just did what [oneTBB's migration guide](https://docs.oneapi.io/versions/latest/onetbb/tbb_userguide/Migration_Guide/Task_Scheduler_Init.html) said.

### DIFF
--- a/opensubdiv/osd/tbbEvaluator.cpp
+++ b/opensubdiv/osd/tbbEvaluator.cpp
@@ -25,7 +25,8 @@
 #include "../osd/tbbEvaluator.h"
 #include "../osd/tbbKernel.h"
 
-#include <tbb/task_scheduler_init.h>
+#define TBB_PREVIEW_GLOBAL_CONTROL true
+#include <tbb/global_control.h>
 
 namespace OpenSubdiv {
 namespace OPENSUBDIV_VERSION {
@@ -215,10 +216,11 @@ TbbEvaluator::Synchronize(void *) {
 /* static */
 void
 TbbEvaluator::SetNumThreads(int numThreads) {
-    if (numThreads == -1) {
-        tbb::task_scheduler_init init;
-    } else {
-        tbb::task_scheduler_init init(numThreads);
+    if (numThreads != -1) {
+		tbb::global_control tbb_global_control(
+			tbb::global_control::max_allowed_parallelism,
+			numThreads
+		);
     }
 }
 
--- a/opensubdiv/osd/tbbEvaluator.h
+++ b/opensubdiv/osd/tbbEvaluator.h
@@ -1226,7 +1226,7 @@ public:
     /// \brief synchronize all asynchronous computation invoked on this device.
     static void Synchronize(void *deviceContext = NULL);
 
-    /// \brief initialize tbb task schedular
+    /// \brief initialize tbb task scheduler
     ///        (optional: client may use tbb::task_scheduler_init)
     ///
     /// @param numThreads      how many threads


davidgyu:
Hi! Would you please submit a signed Contributor License Agreement as described here: [Contributing to OpenSubdiv](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst#contributor-license-agreement) Thank you!

Logarithmus:
@davidgyu would signing the PDF file with PGP and emailing it to you be sufficient?

jilliene:
Filed as internal issue #OSD-359

davidgyu:
A regular signature on the PDF emailed back to us would be best for our process. Thanks!

Logarithmus:
@davidgyu I've just found a bit of free time and sent a signed PDF to you (opensubdiv-cla@pixar.com). It's signed both by hand & with GPG.

davidgyu:
Thank you @Logarithmus !

opensubdiv/osd/tbbEvaluator.cpp
```diff
@@ -215,10 +216,11 @@ TbbEvaluator::Synchronize(void *) {
...
 TbbEvaluator::SetNumThreads(int numThreads) {
-    if (numThreads == -1) {
-        tbb::task_scheduler_init init;
-    } else {
-        tbb::task_scheduler_init init(numThreads);
+    if (numThreads != -1) {
+		tbb::global_control tbb_global_control(
+			tbb::global_control::max_allowed_parallelism,
+			numThreads
+		);
```
anton-potapov, line 223: Changes to TBB behavior made by  `tbb::global_control` only active wile the object is alive.  So having this object as stack variable makes the whole function `TbbEvaluator::SetNumThreads` to has no effects. 

From the [global_control::~global_control specification](https://spec.oneapi.io/versions/latest/elements/oneTBB/source/task_scheduler/scheduling_controls/global_control_cls.html) :
>Destructs a control variable object and ends it’s impact.

e4lam: +1

Isn't the old code here also effectively a no-op too here? This entire `SetNumThreads()` method and references to it should be removed. A library like this has no business trying to control the number of threads used process wide. It should be a decision left to the library user instead.

MoisMoshev:
Hey, I'm trying to build with your branch, but still unable to compile. Getting errors like these from msbuild:
```
"e:\bottleship\code\3p\OpenSubdiv\build\examples\dxViewer\dxViewer.vcxproj" (Rebuild target) (26) ->
(Link target) -> 
  LINK : fatal error LNK1104: cannot open file 'tbb12.lib'
```
and also
```
  E:\bottleship\code\3p\OpenSubdiv\examples\common\glPtexMipmapTexture.cpp(25,10): fatal error C1083: Cannot open include file: 'glLoader.h': No such file or directory
```

I'm trying with the following cmake options:
```
     -DTBB_INCLUDE_DIR='E:/bottleship/code/3p/oneapi-tbb-2021.5.0/include'
     -DTBB_tbb_LIBRARY='e:/bottleship/code/3p/oneapi-tbb-2021.5.0/lib/intel64/vc14/tbb.lib'
     -DTBB_tbb_debug_LIBRARY='e:/bottleship/code/3p/oneapi-tbb-2021.5.0/lib/intel64/vc14/tbb_debug.lib'
```
The libraries are there, but it's looking for tbb12, which I probably need to add to another path? 
Just wondering if my setup is the issue.

svenstaro:
So what about merging this? Sadly this missed 3.5.0 but I'd appreciate it if this went in somewhat soon.

davidgyu:
Thanks for your patience and apologies that this didn't make it into 3.5.0.
We realize this PR has been up for a long time. We'll add a note here when we know when to expect a release with this fix.

e4lam:
I guess you could merge it but really, you might just as well delete the contents of the `SetNumThreads()` method as it's moving from a method that did nothing to a method that does nothing (as was commented >6 months ago).

davidgyu:
Yes, I agree with your earlier comment regarding SetNumThreads()

RBKreckel:
It would be extremely nice if this could be merged just to keep this package and depending packages building. In any case, it seems to me like the discussion about whether the function can be further simplified could be deferred until later.

brad0:
As a OpenBSD package maintainer I came across this. It would be nice to have a proper fix upstream for TBB 2021 support.

davidgyu:
As noted above, we'd like to address the presence of SetNumThreads() in the API along with the other low level fixes.
We have this on deck for our next release (3.6.x) where we'll be able to address API changes (the current 3.5.1 release is for minor bug fixes only). Thanks again for your patience.

davidgyu:
Fixed by #1317 and #1319

e4lam:
Thank you for the update!

PR #1235, opened 2021-10-06
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR.

thomthom:
This is a follow up to my previous PR to make the OSD CMake project be usable as a sub-project. (https://github.com/PixarAnimationStudios/OpenSubdiv/pull/1206)

This time I was working with macOS and found some CMake variables that needed adjusting for `APPLE` specific config.


thomthom:
Side note, in https://github.com/PixarAnimationStudios/OpenSubdiv/pull/1206 I proposed to make some further improvements to the CMake project. Related to that I asked what minimum CMake version this project wanted/needed to support. The response then was that upping the minimum version was being deferred.

Currently I'm running CMake 3.21 and are now seeing deprecation warnings for older CMake configurations:

```
[cmake] CMake Deprecation Warning at ThirdParty/OpenSubdiv/CMakeLists.txt:27 (cmake_minimum_required):
[cmake]   Compatibility with CMake < 2.8.12 will be removed from a future version of
[cmake]   CMake.
[cmake] 
[cmake]   Update the VERSION argument <min> value or use a ...<max> suffix to tell
[cmake]   CMake that the project does not need compatibility with older versions.
```

It's not clear to me how far into the future this compatibility removal might happen, but maybe worth reconsidering the minimum CMake version?

jilliene:
Filed as internal issue #OSD-361

davidgyu:
Thanks! Made a slight additional change to accommodate the bfr library.

PR #1237, opened 2021-10-09
Move cmake_minimum_required to be the first thing in CMakeLists.txt.

thomthom:
Moving to be the very first thing in CMakeLists.txt per CMake's strong recommendation.

> **Note** Call the cmake_minimum_required() command at the beginning of the top-level CMakeLists.txt file even before calling the project() command. It is important to establish version and policy settings before invoking other commands whose behavior they may affect. See also policy CMP0000.

https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

There might be some other `CMAKE_` variables set that ideally should be set before `project`, but those would require some more investigation to identify.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,9 +22,9 @@
 #   language governing permissions and limitations under the Apache License.
 #
 
-project(OpenSubdiv)
-
 cmake_minimum_required(VERSION 2.8.6)
+
+project(OpenSubdiv)
 
 # Turn on folder support
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)


thomthom:
Looking at the CMake files it appear that this project define `CMAKE_` prefixed variables:

https://github.com/PixarAnimationStudios/OpenSubdiv/blob/ff76e0f2dc9c9202b7d2f965f264bfd6f41866d5/CMakeLists.txt#L136-L141

These variables doesn't appear to be standard CMake variables. It gets rather confusing trying to tell when a CMake property is set and when a custom `CMAKE_` variable is set. Maybe worth removing the `CMAKE_` prefix for variables that aren't standard CMake variables?

jilliene:
Filed as internal issue #OSD-363

davidgyu:
Thanks @thomthom

Issue #1236, opened 2021-10-06
Unable to build CPU only binaries on macOS


thomthom:
I have a project where I'm only using osd_cpu.

I configure the build using:

```
mkdir build
cd build
cmake \
  -D NO_EXAMPLES=1 \
  -D NO_TUTORIALS=1 \
  -D NO_REGRESSION=1 \
  -D NO_PTEX=1 \
  -D NO_DOC=1 \
  -D NO_OMP=1 \
  -D NO_TBB=1 \
  -D NO_CUDA=1 \
  -D NO_OPENCL=1 \
  -D NO_CLEW=1 \
  -D NO_OPENGL=1 \
  -D NO_METAL=1 \
  -D NO_DX=1 \
  -D NO_TESTS=1 \
  -D NO_GLTESTS=1 \
  -D NO_GLEW=1 \
  -D NO_GLFW=1 \
  -D NO_GLFW_X11=1 \
  ..
```

This fails to configure with the warning with the output:

```
-- The C compiler identification is AppleClang 12.0.5.12050022
-- The CXX compiler identification is AppleClang 12.0.5.12050022
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Deprecation Warning at CMakeLists.txt:27 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Compiling OpenSubdiv version v3_4_4
-- Using cmake version 3.21.3
-- Configuring done
CMake Error at opensubdiv/CMakeLists.txt:320 (add_library):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:osd_gpu_obj>

  Objects of target "osd_gpu_obj" referenced but no such target exists.


[...]

PR #1238, opened 2021-10-20
Add option to not generate frameworks under macOS. Ref #1236

thomthom:
This doesn't address the issue of the framework failing to build if you try to only build CPU libs, but the option to generate a framework or not seemed different enough to warrant its own PR.

I had a brief look at the Azure setup to see if there was an easy way to add a test configuration of OSD with only CPU libs generated, but that looked to be a larger task.


jilliene:
Filed as internal issue #OSD-364

davidgyu:
Thanks!

PR #1242, opened 2022-04-27
Add CMake Config

theblackunknown:
# Description

This PR adds a CMake Config to OpenSubdiv and expose preprocessor macros for OpenSubdiv enabled features for clients to depend on them.
This allow for a more robust and consistent way to pull `OpenSubdiv` as a dependency in other projects through CMake `find_package` calls.
Moreover, this enhance the experience when using package manager such as [microsoft/vcpkg](https://github.com/microsoft/vcpkg) and alikes.


theblackunknown:
Note that as the minimum required version is 2.8 I have to do some detection.
It would be nice to bump this requirement up to avoid this extra complexity if possible.

thomthom:
Ah, nice. I have an open issue in the vcpkg regarding this: https://github.com/microsoft/vcpkg/issues/20932

Does this PR provide a config that allows to choose what features to use? The vcpkg package had to patch the OSD project to handle this: https://github.com/microsoft/vcpkg/issues/20871

theblackunknown:
> Does this PR provide a config that allows to choose what features to use? The vcpkg package had to patch the OSD project to handle this: microsoft/vcpkg#20871

It seems that Jack already took care of most of the features in microsoft/vcpkg#20895, is this what you are looking for ?

Regarding OpenSubdiv vcpkg port, I plan to improve it with what is not yet supported (e.g. OpengGL backend) later after this PR.

theblackunknown:
> Ah, nice. I have an open issue in the vcpkg regarding this: microsoft/vcpkg#20932

I was not aware of your issue but I feel this we aim for the same goal :)

theblackunknown:
Hey @thomthom , I am working on microsoft/vcpkg#24467 which offers a CMake config in the current vcpkg port while we wait for this PR to move forward.

davidgyu:
Hi, would you submit a [CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) Thanks!

theblackunknown:
> Hi, would you submit a [CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) Thanks!

@davidgyu thanks for the link I did not notice this page before !
Let me run some double check, and then I'll sign it asap

theblackunknown:
> Hi, would you submit a [CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) Thanks!

@davidgyu I have just signed and sent my CLA, let me know when this is good to move forward with this PR

opensubdiv/CMakeLists.txt
```diff
@@ -366,6 +366,43 @@ if (NOT NO_LIB)
         endif()
 
     endif()
+
+    # Expose preprocessor macro to the interface so that client know which feature are enabled
```
theblackunknown, line 381: This portion is useful so that client consuming this any OpenSubdiv library can introspect which feature are enabled directly on the exported CMake targets.

For example, USD would benefit from this as they are checking if header exists and recompute those exact preprocessor macros because they are not exported
cf. https://github.com/PixarAnimationStudios/USD/blob/3b097e3ba8fabf1777a1256e241ea15df83f3065/cmake/modules/FindOpenSubdiv.cmake#L110-L117

theblackunknown:
@davidgyu is there anything new on this PR ?
I am happy to discuss/improve it if need be, otherwise let me know if this is good to merge

theblackunknown:
Gently bumping this PR for any news @davidgyu ?

davidgyu:
Filed as internal issue #OSD-382

davidgyu:
We have a few changes that we're going to be merging ahead of this PR and there may be some additional work (hopefully minor) that will be needed for this PR. We'll post additional notes here when the necessary changes are clear. Thanks!

theblackunknown:
Thanks for the news David, I'll wait for your changes and adapt my PR when they arrived.

davidgyu:
This can be simplified since many of these are not primary artifacts and shouldn't be exported as config targets.

Specifically, we should remove the changes in the **examples**, **regression**, **tutorials**, and **opensubdiv/tools/stringify** CMakeLists.txt files.

Thanks!

thomthom:
And there is a PR open to bump minimum CMake version, which sounds like it would aid in simplifying this PR.

davidgyu:
I've pushed some commits to remove the non-primary export targets and to update the config and target names to be more consistent with other packages in the vfx platform.

davidgyu:
Thank you @theblackunknown this is great to have!

theblackunknown:
This is great !
Thank you @davidgyu for tackling the last bits and your patience !

theblackunknown:
> And there is a PR open to bump minimum CMake version, which sounds like it would aid in simplifying this PR.

I would love too as well, but I think it depends on the minimum CMake version Pixar - and anyone else consuming OpenSubdiv - requires which does not seems to be that recent ?

PR #1244, opened 2022-06-26
OpenGL Compute Evaluator: cache active program

diekev:
Client applications may keep track of the active shader program so as to
not unnecessarily bind and unbind the same shader consecutively as
switching shaders is costly.

The current behavior of the OpenGL compute evaluator is to set the
active shader to 0 (zero) after running its program(s) which might
interfere with the aforementioned client applications' behavior,
leading to bugs.

Instead of unsetting any previously set shader program, cache the
current program before using the evaluator program, and reset the active
program to the previous one when done.

In Blender, this lead to some rendering artifacts. See
https://developer.blender.org/D15064 for more details on it.


davidgyu:
Filed as internal issue #OSD-383

davidgyu:
Thanks!

PR #1246, opened 2022-07-18
Updated macOS azure pipeline agent

davidgyu:
The current macOS-10.14 agent has been retired. This updates to use macOS-10.15 which matches the config for the USD repository. But we may update this further to macOS-11 since the macOS-10.15 agent has recently been marked deprecated.


davidgyu:
Closing to reopen on the dev branch

PR #1247, opened 2022-07-18
Updated macOS azure pipelines agent

davidgyu:
The current macOS-10.14 agent has been retired. This updates to use macOS-10.15 which matches the config for the USD repository. We may update this further to macOS-11 since the macOS-10.15 agent recently has been marked deprecated.


davidgyu:
Filed as internal issue #OSD-370

PR #1253, opened 2022-08-30
Suppression of compiler warnings from GCC

barfowl:
These changes address compiler warnings from GCC -- most notably those arising from more recent versions of GCC that trigger warnings from OpenSubdiv's public header files, and so interfere with the compilation of client code.

The changes consist of two commits:  the first addresses warnings arising from OpenSubdiv's use of -Wall and -Wextra, while the second addresses some of the additional warnings arising from -Wshadow and -Wconversion.

Warnings are addressed in the libraries, tutorials and regression tests, but not the examples or their common infrastructure.

### DIFF
--- a/examples/common/stopwatch.h
+++ b/examples/common/stopwatch.h
@@ -43,14 +43,15 @@ public:
     void Start() {
         struct timeval l_rtime;
         gettimeofday(&l_rtime,0);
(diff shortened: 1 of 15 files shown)


davidgyu:
Thanks, Barry!

PR #1255, opened 2022-09-02
Added more inspection methods to Far::TopologyLevel

barfowl:
Changes here include additional methods for Far::TopologyLevel to provide more information about the vertices and edges of a level.

These include tests on whether a vertex is a corner (only one incident face), whether the valence of a vertex is regular (with respect to the assigned subdivision scheme), and whether a vertex or edge is semi- or inf-sharp.

Some of these tests make use of internally assigned bits for the vertex or edge and so will be a little more efficient than combining other existing tests.

These changes also cover the additions proposed in #1227.


davidgyu:
Thanks, Barry!

Issue #1269, opened 2022-09-14
Compile errors while building with Ninja Release 3_4_4


roquo:
Hello,

I tried to build OpenSubdiv with Ninja. Unfortunately, I get a lot of warnings that are treated as errors and the build fails.

My setup:
```
Microsoft Visual Studio Professional LTSC 2022 17.2.3
Windows Kits SDK 10.0.19041.0
CMake 3.20.5
Ninja 1.10.2
```

The commands I run:
```
git clone https://github.com/PixarAnimationStudios/OpenSubdiv
cd OpenSubdiv
mkdir build
mkdir install
cd build
set INCLUDE=C:/Program Files/Microsoft Visual Studio/2022/17.2/VC/Tools/MSVC/14.32.31326/ATLMFC/include;C:/Program Files/Microsoft Visual Studio/2022/17.2/VC/Tools/MSVC/14.32.31326/include;C:/Program Files/Windows Kits/NETFXSDK/4.8/include/um;C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt
set LIB=C:/Program Files/Microsoft Visual Studio/2022/17.2//VC/Tools/MSVC\14.32.31326\lib\x64;C:/Program Files/Microsoft Visual Studio/2022/17.2/VC/Tools/MSVC/14.32.31326/ATLMFC/lib/x64
cmake -G Ninja .. -DNO_EXAMPLES=1 -DNO_TUTORIALS=1 -DNO_REGRESSION=1 -DNO_PTEX=1 -DNO_DOC=1 -DNO_OMP=1 -DNO_TBB=1 -DNO_CUDA=1 -DNO_OPENCL=1 -DNO_CLEW=1 -DNO_OPENGL=1 -DNO_METAL=1 -DNO_DX=1 -DNO_TESTS=1 -DNO_GLTESTS=1 -DNO_GLEW=1 -DNO_GLFW=1 -DNO_GLFW_X11=1 -DBUILD_SHARED_LIBS=0 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=..\install
ninja install
```

Output after running CMake:
```
-- The C compiler identification is MSVC 19.32.31329.0
-- The CXX compiler identification is MSVC 19.32.31329.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
[...]

PR #1273, opened 2022-09-16
added additional compiler flags to remove warnings created on ninja build

roquo:
Issue: [Compile errors while building with Ninja Release 3_4_4](https://github.com/PixarAnimationStudios/OpenSubdiv/issues/1269)


davidgyu:
Filed as internal issue #OSD-391

davidgyu:
Thanks for taking a look at this. The direction we'd like to take is to have a consistent set of compiler flags for MSVC regardless of generator, i.e. not ninja specific.

roquo:
@davidgyu So does that mean the ignored warnings should be added directly for MSVC? I am not sure why the warnings only come up with the Ninja generator and not with Visual Studio. It would be a great help for me to directly support ninja without applying patches. Thanks! I am going to update the pull request

PR #1275, opened 2022-09-27
Avoid referring to std::vector<T> members when T is incomplete.

Cory-Kramer:
Avoid referring to std::vector<T> members when T is incomplete.

This is not legal according to the C++ standard, and causes build errors when using --std=c++20

### DIFF
--- a/opensubdiv/far/patchTable.cpp
+++ b/opensubdiv/far/patchTable.cpp
@@ -125,23 +125,6 @@ PatchTable::~PatchTable() {
 //
 // PatchArrays
 //
-struct PatchTable::PatchArray {
-
-    PatchArray(PatchDescriptor d, int np, Index v, Index p, Index qo) :
-            desc(d), numPatches(np), vertIndex(v),
-                patchIndex(p), quadOffsetIndex (qo) { }
-
-    void print() const;
-
-    PatchDescriptor desc;  // type of patches in the array
-
-    int numPatches;        // number of patches in the array
-
-    Index vertIndex,       // index to the first control vertex
-          patchIndex,      // absolute index of the first patch in the array
-          quadOffsetIndex; // index of the first quad offset entry
-
-};
 
 // debug helper
 void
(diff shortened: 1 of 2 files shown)


davidgyu:
Hi, would you submit a [CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) Thanks!

Cory-Kramer:
Hi David, I am making this change on behalf of Google. Do you already have a Corporate CLA filed for Google? From what I can tell it should have previously been filled out either on behalf of Google or perhaps Waymo.

Cory-Kramer:
Hi David, were you able to verify that Google has a CLA submitted?

davidgyu:
@Cory-Kramer we're checking on our side and we'll follow up here when we know more. Thanks!

davidgyu:
@Cory-Kramer we don't have a corporate CLA from Google or Waymo in our files. Would you be able to kick off the process to get that going? Thanks!

Cory-Kramer:
Sure thing, I'll get that process started on our end. I'll update when I have more details.

Cory-Kramer:
Hi @davidgyu I have forwarded our Corporate CLA to opensubdiv-cla@pixar.com please let me know if you need anything else. Thanks!

davidgyu:
Thanks, @Cory-Kramer! We've received the Corporate CLA and will begin reviewing the submission.

davidgyu:
Filed as internal issue #OSD-392

davidgyu:
Can you tell us more about how you are encountering this error, i.e. compiler version, compiler flags, os, etc?

While there are clearly opportunities to improve the declaration of Far::PatchTable we want to understand the problem better.
 
We've tried some simple experiments with specifying --std=c++20 for various recent compilers and haven't been able to reproduce the errors yet.

Cory-Kramer:
This restriction comes from §22.3.11.1 of the standard library ([link](https://timsong-cpp.github.io/cppwp/n4868/vector#overview-4)):
>An incomplete type `T` may be used when instantiating `vector` if the allocator meets the allocator completeness requirements. `T` shall be complete before any member of the resulting specialization of `vector` is referenced.

To reproduce see the following example ([working demo](https://godbolt.org/z/sGK876EdK))
```cc
#include <vector>

struct Data;  // forward declaration (incomplete type)

struct Container {
    std::size_t DoSomething() const {
        return data.size();
    }
    std::vector<Data> data;  // usage of incomplete type in vector member
};

struct Data {};  // full type definition

int main() {
    Container c;
}
```
Compiling this example using clang with `-O3 -std=c++20` results in the following build error
```cc
In file included from <source>:1:
In file included from /opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/13.0.0/../../../../include/c++/13.0.0/vector:66:
/opt/compiler-explorer/gcc-snapshot/lib/gcc/x86_64-linux-gnu/13.0.0/../../../../include/c++/13.0.0/bits/stl_vector.h:988:50: error: arithmetic on a pointer to an incomplete type 'Data'
      { return size_type(this->_M_impl._M_finish - this->_M_impl._M_start); }
                         ~~~~~~~~~~~~~~~~~~~~~~~ ^
<source>:7:21: note: in instantiation of member function 'std::vector<Data>::size' requested here
        return data.size();
                    ^
<source>:3:8: note: forward declaration of 'Data'
struct Data;  // forward declaration (incomplete type)
       ^
1 error generated.
Compiler returned: 1
```

davidgyu:
Thank you for the additional context!

Cory-Kramer:
Hi @davidgyu do you have an estimated date for the next `release` tag?

davidgyu:
Our current plan is to finalize development by the end of this month for the next patch release. Our release process includes a week of RC testing so this fix should be available in the release branch within the next month.

PR #1298, opened 2023-03-10
Fixed glFVarViewer display of singlecrease patches

davidgyu:


davidgyu:
This change triggers some apparent broken behavior in Mesa drivers (observed on `Mesa Intel(R) UHD Graphics (CML GT2) 4.6 (Compatibility Profile) Mesa 22.0.5` and `Mesa Intel(R) Graphics (RPL-S) 4.6 (Compatibility Profile) Mesa 22.2.5`)

Similar broken behavior can be triggered on the same drivers using the existing glViewer built from the release branch.

Investigating to see if there's a way we can work around these driver bugs...

PR #1301, opened 2023-05-25
Fixed Bfr::Surface construction bug for rare topological case

barfowl:
This change fixes a bug in the Bfr::SurfaceFactory that occurs for faces with a rare combination of topological features:  a valence-2 interior vertex, a boundary edge and at least one explicitly assigned semi-sharp edge. The presence of a valence-2 interior vertex creates the possibility of a self-overlapping face neighborhood, leading to a separate branch of code for such cases (where the bug occurs), so the change is isolated.

In addition to fixing the bug, a revision to one of the test shapes is included that now captures the combination of features that triggered it.


davidgyu:
Filed as internal issue #OSD-413

davidgyu:
Thanks, Barry!

PR #1306, opened 2023-06-26
hbr/mesh.h: Fix warning, numfaces written but not read.

waywardmonkeys:
While it was iterating over the faces and calculating counts, `numfaces` was being incremented, but then `nfaces` was used. These would differ if there was a null face.

Found when using roughly current compilers for emscripten and a WebAssembly build.

### DIFF
--- a/opensubdiv/hbr/mesh.h
+++ b/opensubdiv/hbr/mesh.h
@@ -899,8 +899,8 @@ HbrMesh<T>::PrintStats(std::ostream &out) {
             sumsides += f->GetNumVertices();
         }
     }
-    out << "Mesh has " << nfaces << " faces\n";
-    out << "Average sidedness " << (float) sumsides / nfaces << "\n";
+    out << "Mesh has " << numfaces << " faces\n";
+    out << "Average sidedness " << (float) sumsides / numfaces << "\n";
 }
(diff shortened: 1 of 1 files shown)


davidgyu:
Filed as internal issue #0SD-424

davidgyu:
As noted here: [3.x Hbr Overview](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/hbr_overview.rst) and here [3.5 Release Notes](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/release_notes.rst#release-3-5) the code in `opensubdiv/hbr` is deprecated and slated for removal at some point in the future. We're reluctant to merge changes into this part of the codebase.
We are looking into adding a `NO_HBR` option which hopefully would help you out here.
Thanks!

waywardmonkeys:
I hadn't seen that. Would an external PR to add such an option be of interest or will you just do it internally?

davidgyu:
A PR would be welcome but be forewarned that there are currently dependencies on hbr in the regression tests which complicates implementing such a NO_HBR option.

davidgyu:
Marking this closed per discussion above.

PR #1315, opened 2023-08-12
allow absolute GNU install dirs

chayleaf:
`CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` are allowed to be absolute [as per CMake docs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), so they shouldn't be concatenated to absolute paths. Instead, use `CMAKE_INSTALL_FULL_LIBDIR` and `CMAKE_INSTALL_FULL_INCLUDEDIR`, which are known to be absolute.

Absolute `CMAKE_INSTALL_*DIR` paths are used at least on NixOS. This is because NixOS packages are allowed to have multiple outputs (e.g. put docs in one dir, headers and other files for development to another, the binaries to a third one), and NixOS isn't FHS-compliant (There's no `/usr`, each package output has its own directory instead).

GNU Guix is a similar distro, but I'm not sure how CMake packages are built there. Either way, it's good to fully conform to the spec.

### DIFF
--- a/opensubdiv-config.cmake.in
+++ b/opensubdiv-config.cmake.in
@@ -1,6 +1,6 @@
 @PACKAGE_INIT@
 
-set_and_check(OpenSubdiv_INCLUDE_DIR "${PACKAGE_PREFIX_DIR}/@CMAKE_INSTALL_INCLUDEDIR@")
-set_and_check(OpenSubdiv_LIB_DIR "${PACKAGE_PREFIX_DIR}/@CMAKE_INSTALL_LIBDIR@")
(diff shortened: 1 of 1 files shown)


davidgyu:
Filed as internal issue #OSD-430

davidgyu:
Hi, @chayleaf would you submit a [CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) Thanks!

chayleaf:
Just did that

Issue #1307, opened 2023-06-30
Deprecated  tbb::task_scheduler_init


marza-sergey:
`tbb::task_scheduler_init` was deprecated and gone in newest releases of oneTBB.
https://github.com/PixarAnimationStudios/OpenSubdiv/blob/8ffa2b6566be10209529d7a0d1db02a0796b160c/opensubdiv/osd/tbbEvaluator.cpp#L28

davidgyu:
Filed as internal issue #OSD-425

davidgyu:
There's some additional discussion of this in PR #1234

davidgyu:
Fixed by #1317

PR #1317, opened 2023-09-07
Marked deprecated Osd::*Evaluator::SetNumThreads()

### LINKED ISSUE #1211: Adding support for oneTBB

davidgyu:
The existing implementation in Osd::TbbEvaluator has never been implemented correctly and that implementation is incompatible with the oneTBB API.

These methods are now marked deprecated in the two places where defined in Osd::TbbEvaluator and Osd::OmpEvaluator, but for consistency we preserve the existing implementation when building with older versions of tbb.

We've already removed all use of these methods from the tests and examples.

Fixes #1211 #1234 #1307


davidgyu:
These changes have been tested on Windows, Linux, and macOS using TBB 2018 Update 6, TBB 2020 Update 3 and oneTBB 2021.10.0

davidgyu:

PR #1319, opened 2023-09-11
Find TBB using TBB's CMake config

### LINKED ISSUE #1211: Adding support for oneTBB

davidgyu:
Updated the CMake build to locate TBB using TBB's CMake config and deleted the local cmake/FindTBB.cmake module.

While the source code remains compatible with earlier versions of TBB, updated the minimum TBB version from 4.4 to 2018 since that is the earliest TBB release to include a CMake config.  This is the version of TBB specified for the CY2019 VFX Reference Platform.

Tested on Windows, Linux, and macOS using TBB 2018 Update 6, TBB 2020 Update 3, and oneTBB 2021.10.0

Fixes #1211 #1234


davidgyu:

Issue #1153, opened 2019-07-27
Vulkan back-end?


cesss:
I see there's no Vulkan back-end, but I believe it would be a great idea in many aspects:

- It's an API with concepts targeted to current gfx hardware designs, closer to Metal than to OpenGL.
- It's available on Windows, Linux, and MacOS (on MacOS you need the MoltenVK lib, which works on top of Metal, but I tried it and it works really great).
- Performance of OpenSubdiv should be better than with OpenGL.
- Gfx developers are progressively turning their eyes to Vulkan (just to put an example, even [ImGui](https://github.com/ocornut/imgui) has a Vulkan back-end).

I'm going to start Vulkan back-ends for my projects in the upcoming months, so I'd like that OpenSubdiv will be supported on them too.

jilliene:
Filed as internal issue #OSD-298

davidgyu:
Yes, we agree. We’ve been tracking progress on Vulkan (and MoltenVK) since the beginning and we are very excited about the opportunities there, but it hasn’t yet been a Studio priority. Much of the existing GLSL or HLSL shader can be used directly which simplifies the next steps of developing a Vulkan back-end implementation.

jeroenbakker-atmind:
Hi, 

Here at Blender we are migrating towards Vulkan. The challenge is that we expect to use the OpenGL evaluator in Blender 3.1 (https://developer.blender.org/D12406) and when we introduce Vulkan we can't/won't want to use the OpenGL evaluator. Is there any news on this topic?

In our current plans the Vulkan migration will speed up on our side and are planning to have a Vulkan running in Blender within the next year. Perhaps we can see how to move forward with modern backends.

Regards,
Jeroen Bakker

davidgyu:
Thank you for including that link to your work in progress. I haven't had a chance to look through your code yet, but the comments in that thread raise good questions and hopefully we can help you get things working well. We are definitely interested in having Blender take good advantage of OpenSubdiv.

We've been getting some good results using Vulkan and OpenSubdiv together including patch evaluation.

bsavery:
Hi @davidgyu Any update on Vulkan support?

RedMser:
Seems like this issue was supposed to be closed by #1320 but it was wrongly tagged?

[...]

PR #1320, opened 2023-09-13
Improved osd patch drawing compatibility

davidgyu:
Submitted as two commits for a cleaner commit history.

--

Reorganized legacy patch drawing shader source
    
Relocated from PatchCommon to PatchLegacy several aspects of the
shader source which can cause problems with typical use cases.
Specifically, things like resource bindings, input assembler and
interstage declarations are best left to client code.
    
These are not removed, just relocated and remain available for
backward compatibility. Updated the GLSL, HLSL, and MSL source.

--

Added GetPatchDrawingShaderSource()
    
This is a new method for GLSL, HLSL, and MSL which returns patch
drawing shader source which excludes legacy shader source aspects.
    
This improves portability and compatibility and may also improve
shader compile times since the resulting shader source strings
are smaller without the legacy shader source aspects.
    
Tested with Vulkan and DX12 in addition to GL, DX11, and Metal.

Fixes #938 #1153

### DIFF
--- a/opensubdiv/osd/glslPatchCommon.glsl
+++ b/opensubdiv/osd/glslPatchCommon.glsl
@@ -22,101 +22,11 @@
 //   language governing permissions and limitations under the Apache License.
 //
 
-//
-// typical shader composition ordering (see glDrawRegistry:_CompileShader)
-//
-//
-// - glsl version string  (#version 430)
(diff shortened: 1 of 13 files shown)


davidgyu:

PR #1321, opened 2023-09-13
Improved osd patch basis compatibility

davidgyu:
Submitted as two commits for a cleaner commit history.

--

Reorganized osd/patchBasis and osd/patchBasisTypes
    
Folded osd/patchBasisCommon.h and osd/patchBasisCommonEval.h into the
single header file osd/patchBasis.h and renamed osd/PatchBasisCommonTypes.h
to osd/patchBasisTypes.h
    
Moved these into the OpenSubdiv::Osd namespace when compiled for C++ and
added a brief comment describing the intended use cases for this code.

--

Installed osd/patchBasis.h and osd/patchBasisTypes.h
    
Providing these as public headers in opensubdiv/osd allows client code
to implement evaluation kernels in their own TBB, CUDA, OMP, GCD, etc.
code without needing an implementation within the osd codebase.
    
Tested with C++, CUDA, and oneTBB client kernels.

Fixes #938 #1153


davidgyu:

PR #1322, opened 2023-09-13
Fixed cmake/FindOpenCL.cmake version parsing

davidgyu:
Current versions of the OpenCL headers define several version macros and we want to make sure to parse just the main version macro.

This is a simple near-term fix, since we'd like to move to the standard CMake FindOpenCL module and delete this custom module.


davidgyu:
Reviewed internally.

PR #1331, opened 2024-03-25
Optimize PNGs to save 5MB+ losslessly

C0rn3j:
Exact commands as per https://rys.rs/optimization

Shouldn't break anything, file savings are all lossless with no effect on quality.

I did also correct a PNG that had a .jpg extension and edited its associated .rst mention.


davidgyu:
Filed as internal issue #OSD-440

davidgyu:
We'll need a signed [Contributor License Agreement](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) from you for this PR. Thanks!

C0rn3j:
CLA signed

davidgyu:
Many of these image files have been autogenerated by external tools and we'll have to consider whether we want to take on this additional optimization step. The overall note that many of these image files take up more space than is needed is a great observation and something that we should definitely consider as we add or update content in the future.

C0rn3j:
It certainly would be a good idea to implement a solution to prevent pushing new files with the same issue.

One solution that comes to mind: 
* YAML file that's a list of `filepath: sha256sum`
* pre-commit hook that verifies that the committed files with specific extensions (ideally check mimetypes via `file` or similar instead) are on the list with the matching sum, and if not, run the optimizer locally and push a new checksum
* CI check that verifies the checksums too

There are possibly better ways to do this, it's just an idea I was pondering while sending out patches like these.

Though, even without any future-proof solution being implemented, pulling in this one-time run would help a fair bit, as it targets images across the last 12+ years.  
For example, my distribution ships OpenSubdiv along with the documentation, so the resulting installation size is needlessly large without it.

PR #1333, opened 2024-05-15
Updated build_osd.py for recent versions of Python 3.x

### LINKED ISSUE #1330: Azure pipelines CI failing on macOS
As mentioned in #1329 CI builds are currently failing on macOS because of changes to the config of the Microsoft hosted agent.

We've fixed a similar issue in the OpenUSD build script and can apply a similar fix here.

We might also take this opportunity to update to the `macOS-12` vmImage from the currently specified `macOS-11` vmImage.

```
/bin/bash --noprofile --norc /Users/runner/work/_temp/4a641c5b-5932-47b6-9975-7d6d26e3535f.sh
/Users/runner/work/1/s/build_scripts/build_osd.py:122: SyntaxWarning: invalid escape sequence '\d'
  "(\d+).(\d+)",
Traceback (most recent call last):
  File "/Users/runner/work/1/s/build_scripts/build_osd.py", line 26, in <module>
    from distutils.spawn import find_executable
ModuleNotFoundError: No module named 'distutils'
```

davidgyu:
Removed use of distutils for Python 3.x since distutils was removed from Python 3.12.x after being deprecated for a long 
time.

This has been tested with both Python2 and Python3 and restores support for macOS-11 CI builds.

Since Python2 is deprecated we'll submit a follow-up change to remove the remaining vestiges of Python2 support.

Fixes #1330

### DIFF
--- a/build_scripts/build_osd.py
+++ b/build_scripts/build_osd.py
@@ -23,13 +23,10 @@
 #
 from __future__ import print_function
 
-from distutils.spawn import find_executable
-
 import argparse
 import codecs
 import contextlib
 import datetime
-import distutils
 import fnmatch
 import glob
 import locale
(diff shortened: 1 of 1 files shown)


davidgyu:
Reviewed internally.

PR #1337, opened 2024-06-17
Updated azure pipelines for macOS

davidgyu:
- macOS 12 Monterey (updated from macOS 11, unsupported after 6/28/2024)


davidgyu:
Reviewed internally.

PR #1342, opened 2024-12-04
Updated azure pipelines for macOS

davidgyu:
- macOS 13 Ventura (updated from macOS 12 unsupported after 12/03/2024)


davidgyu:
Reviewed internally.

PR #1343, opened 2024-12-13
Fix address sanitizer warnings with stencil table casting

brechtvl:
For compatibility, `StencilTableReal<float>` is cast to `StencilTable`, which has the same data layout but it's not actually an instance of that class. And so using `static_cast` is not strictly correct. Use `reinterpret_cast` instead.


davidgyu:
Hi Brecht! Would you mind submitting an [OpenSubdiv CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) which is distinct from the OpenUSD CLA we already have on file. Thanks!

brechtvl:
Ah, I didn't realize. Sent now.

davidgyu:
Filed as internal issue #OSD-451

PR #1349, opened 2025-03-11
Updated CMake minimum version to 3.14

davidgyu:
We typically test with more recent versions of CMake and we will update this further, but this is the minimum version which allows us to continue improving support for building for embedded systems like iOS.


davidgyu:
Filed as internal issue #OSD-439

PR #1366, opened 2025-09-19
Improve Bfr::Paramaterization conversions to sub-face coordinates

barfowl:
The Bfr::Parameterization methods for non-quads that convert coordinates between the different sub-face representations do not currently handle values that are significantly out-of-bounds very well (this includes ConvertCoordToSubFace() and its normalized variant).

For a face with N sub-faces, given an arbitrary 2D coordinate -- potentially well outside the domain of any of its N sub-faces -- the simple formulae for the sub-face index and coordinate can yield an index outside the expected [0,N-1] range. This can lead to problems when dealing directly with the returned index, and can also cause Bfr::Surface evaluation methods to crash, since they use it internally.

The change here ensures that the conversions always produce a sub-face index in the desired [0,N-1] range. The resulting coordinate is extrapolated relative to the chosen sub-face and will be outside its domain, as expected. Results when using such out-of-bounds coordinates for evaluation remain undefined, as documented.

### DIFF
--- a/opensubdiv/bfr/parameterization.cpp
+++ b/opensubdiv/bfr/parameterization.cpp
@@ -150,31 +150,30 @@ Parameterization::convertCoordToSubFace(bool normalized,
 
     assert(HasSubFaces());
 
-    int uTile = (int) uvCoord[0];
-    int vTile = (int) uvCoord[1];
+    //  Include an interval around the domain when identifying the subface:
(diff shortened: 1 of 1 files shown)


davidgyu:
Filed as internal issue #OSD-464

davidgyu:
Also, filed a separate internal issues to track the macOS CI build failures we have started hitting recently #OSD-465

PR #1367, opened 2025-09-25
Add CMake Option to Disable Static Libs

danrbailey:
In some cases, building only the shared libraries is desirable. The default behavior remains to build both shared and static libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -344,6 +344,11 @@ option(OSD_PATCH_SHADER_SOURCE_MSL "MSL Patch Shader Source" OFF)
 option(OPENSUBDIV_GREGORY_EVAL_TRUE_DERIVATIVES "Enable true derivative evaluation for Gregory basis patches" OFF)
 
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(BUILD_STATIC_LIBS "Build static libraries" ON)
+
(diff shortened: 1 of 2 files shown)


davidgyu:
Filed as internal issue #OSD-466

davidgyu:
@danrbailey could you make sure to submit an [OpenSubdiv Contributor License Agreement (CLA)](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst). Thanks!

davidgyu:
Hey @danrbailey checking again to see if you could submit an [individual or corporate CLA](https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/documentation/contributing.rst) to allow us to consider merging these PRs. Thanks! -David

nvmkuruc:
I wanted to ask about this approach.  `BUILD_SHARED_LIBS` is a first class documented cmake variable (https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) while `BUILD_STATIC_LIBS` is not.

To support the request of being able to build `static` without `shared`, would it be better to follow the guidance in `BUILD_SHARED_LIBS` and produce either shared or static libraries and _never_ both?

This would align to standard cmake conventions and let the same target identifier be used in both static and shared builds.

PR #1368, opened 2025-09-25
Prefer TBB CMake Config Files (OneTBB)

danrbailey:
This makes two changes:

* Prefers to use the TBB CMake config files that ship with OneTBB (if available).
* Add a new `REQUIRE_TBB` so that the build will fail if TBB is not found. 
 
The second change is potentially more contentious, but we ran into an issue where TBB wasn't being found and quietly being switched off as a result. The warning in the output log was not noticed and it wasn't until later that it was discovered that this mis-configuration meant that there were no TBB OpenSubdiv symbols present. I am not sure if this pattern is one that is amenable, but it would be nice to have a convention to prevent the build from continuing when a dependency is not found. The default behavior remains to quietly drop the dependency.


davidgyu:
Filed as internal issue #OSD-467

davidgyu:
We removed OpenSubdiv's FindTBB module a while ago in #1319, but agree that specifying CONFIG mode here makes this even more explicit. We merged a simpler fix as #1374.

davidgyu:
Marking this closed since we've implemented the simpler fix to always require CONFIG mode when finding TBB.

Agree that this is the preferred way to locate TBB dependencies. We've previously tested that this works all of the way back through TBB 2018, though we hope and expect that most users are now using oneTBB.

Your idea to be able to upgrade optional features to required is interesting but we'll defer that for now.

Thanks!

PR #1369, opened 2025-09-25
Adapt OpenSubDiv to use CudaToolkit instead of Cuda

danrbailey:
This is the more modern way of finding the Cuda Toolkit which uses target-based linking. It also takes advantage of Cuda being a first-class language in CMake. The legacy `find_package(Cuda)` mechanism used in OpenSubdiv is now deprecated. We have found this to work more reliably with later versions of Cuda on Linux.

This does however require a bump from 3.14 to 3.17 as the minimum CMake version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@
 #   language governing permissions and limitations under the Apache License.
 #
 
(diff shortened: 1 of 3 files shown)


davidgyu:
Filed as internal issue #OSD-468

PR #1370, opened 2025-09-25
Only enable the install rpath if CMAKE_SKIP_RPATH is not set

danrbailey:


davidgyu:
Filed as internal issue #OSD-469

PR #1372, opened 2025-10-12
Improved minimal Bfr::Tessellation patterns for quads

barfowl:
These changes improve Bfr::Tessellation patterns for quads when all of the tessellation rates in one of the U or V directions are the minimum 1.

As noted by one of our rendering clients (thanks @elalish), the addition of a central point in the face and the resulting triangle fan can be highly objectionable in these cases:

> before

The changes here avoid adding the center point when the tessellation rate is 1 across an entire direction. Differing rates between opposing edges in the other direction are accommodated by generating a combination of quad and triangle facets that span the entire face (or just triangles if quad preservation is not enabled):

> after

The same code that is used to generate facets between the outer and inner edges of a non-uniform tessellation was adapted for use here, so it retains the same behavior regarding quad preservation and symmetry.

For those wanting to preserve the existing behavior, appending inner rates of 2 in these cases will force the creation of the center point.

### DIFF
--- a/opensubdiv/bfr/tessellation.cpp
+++ b/opensubdiv/bfr/tessellation.cpp
@@ -434,16 +434,15 @@ namespace {
         //  over the integer range [0 .. M*N] where M and N are the resolution
         //  (number of edges) of the inner and outer rings respectively.
         //
-        //  Note that the current implementation expects the faces at the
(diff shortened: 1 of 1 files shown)


davidgyu:
Filed as internal issue #OSD-471

PR #1374, opened 2025-10-24
Updated to require config mode when finding TBB

davidgyu:
For modern versions of TBB and oneTBB we want to use CONFIG mode when finding TBB package components. While we have removed the local FindTBB.cmake module we want to guard against accidentally picking up some other find module in the module search path.


davidgyu:
Filed as internal issue #OSD-474

Issue #1379, opened 2026-01-15
Double free in Vtr::internal::StackBuffer when shrinking then destroying SurfaceData


hgarrereyn:
Hi, there is a potential bug in `SurfaceData` reachable when resizing and then deleting.

This bug was reproduced on https://github.com/PixarAnimationStudios/OpenSubdiv/commit/9dab8a47bfbb1388ec8388fe61f5f916e6123f38.

### Description

What crashes
- SurfaceData stores its control-vertex indices in a Vtr::internal::StackBuffer<Index,20,true> (inline capacity 20).
- Calling SurfaceData::resizeCVs(28) grows the StackBuffer above its inline capacity and allocates dynamic storage.
- Calling SurfaceData::resizeCVs(0) shrinks it below the inline capacity. StackBuffer::SetSize() calls deallocate(), which frees the dynamic block and switches back to the inline buffer.
- Deallocate() does not null _dynamicData, and the destructor of StackBuffer then calls deallocate() again unconditionally (or equivalently without guarding on current storage), leading to a double free of the same pointer.
- AddressSanitizer reports a double-free at vtr/stackBuffer.h:100 in both the original reproducer and the minimized testcase.

Note that removing the first resize (`s->resizeCVs(28)`) no longer crashes with a double free.

### POC

The following testcase demonstrates the bug:

**testcase.cpp**
```cpp
#include "/fuzz/install/include/opensubdiv/bfr/surfaceData.h"

int main(){
  using namespace OpenSubdiv::OPENSUBDIV_VERSION::Bfr::internal;
  SurfaceData *s = new SurfaceData();
  // Grow beyond inline capacity (SIZE = 20), which allocates dynamically
  s->resizeCVs(28);
  // Shrink to <= inline capacity, which frees the dynamic storage
  s->resizeCVs(0);
  // Destructor frees again: double-free
  delete s;
  return 0;
}

```

**stdout**
```text

```

**stderr**
```text
=================================================================
==1==ERROR: AddressSanitizer: attempting double-free on 0x50b000000040 in thread T0:
    #0 0x55f8cecb398d in operator delete(void*) (/fuzz/test+0x10598d) (BuildId: 22f0dd75076291483718a1eb9bd66c7bb72c2319)
[...]

PR #1384, opened 2026-08-13
Fix potential double free in Bfr::Surface::operator=()

barfowl:
As noted in #1379, an oversight in Vtr::internal::StackBuffer might lead to a double free when used by Bfr::Surface to resize the contributing number of CVs.

In most practical usage, i.e. re-population of a Surface via the SurfaceFactory, this is not possible, as the factory never resizes the number of CVs to zero. But use of Surface::operator=() can trigger the noted failure. When the instance on the left-hand side has had memory dynamically allocated, and the instance on the right-hand size was default-initialized (with 0 CVs), the left-hand side will be resized to zero -- triggering the double free.

The change here fixes the bug in StackBuffer with the missing pointer assignment. Additional use of "0" for pointer assignment in the file was also updated to make use of "nullptr".

### DIFF
--- a/opensubdiv/vtr/stackBuffer.h
+++ b/opensubdiv/vtr/stackBuffer.h
@@ -98,6 +98,7 @@ inline void
 StackBuffer<TYPE,SIZE,POD_TYPE>::deallocate() {
 
     ::operator delete(_dynamicData);
+    _dynamicData = nullptr;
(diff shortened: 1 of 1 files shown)


davidgyu:
Filed as internal issue #OSD-481

PR #1385, opened 2026-08-13
Add const qualifier to Bfr::Tessellation transform methods

barfowl:
Changes here correct an oversight in the declarations of public methods of Bfr::Tessellation. The three methods that transform the coordinate indices of a given tessellation are now declared "const" as originally intended.


davidgyu:
Filed as internal issue #OSD-482

PR #1386, opened 2026-08-13
Improve Bfr::Tessellation's quad-preserving patterns

barfowl:
These changes improve the tessellation patterns of Bfr::Tessellation when Options::PreserveQuads is enabled for a non-uniform tessellation. The improved patterns are supported for both quad and non-quad faces (Surfaces) of a mesh whenever Options::PreserveQuads is applicable, i.e. when using a quad-based subdivision scheme.

A second commit here updates images used in the documentation of the Bfr interface -- some of which show the improvements to quad-preserving patterns (for both quads and non-quads), and some of which replace quad-preserving patterns with triangular (in keeping with hardware tessellation comparisons).

Examples below indicate the current state followed by the improved patterns in both general and exceptional cases...
<br>
A rendering client noted that the aspect ratio of the quads along an edge can often be undesirably skewed with non-uniform tessellation rates, as illustrated here along the top row of facets:

current-split

This occurs with non-uniform tessellations as potential quads at the corners are "split" when the rates around the corner are not uniform. This splitting of the corner quads was done to isolate the strips of facets along an edge from tessellation rates on adjoining edges -- simplifying the generation of facets for each edge.  

The improvement here is not to split the corner quads. Instead, when possible the quad facet at the corner is shared by both edges (as is currently done when the rates around the corner are uniform), e.g.:

improved

This reduces the skew seen in the original example. (The presence of three triangles at the center of the edge instead of a triangle and a quad is an intentional property of the patterns that preserves symmetry.)

In some cases, however, the potential corner quads cannot (or should not) be preserved and so the corner quad is split. In such cases, the corner is split into two (or more) triangles -- preserving the improved strip of quads along the rest of the edge, e.g.:

improved-split

The example here shows the left edge with a rate of 1 forcing the split of potential corner quads. Corners may also be split when the rate is not 1 if the difference in rates around the corner (both outer and inner) would produce a quad with a poor aspect ratio relative to its neighbors.
<br>
Note:  A side effect of these changes does have a minor impact on tessellations that do not use the quad preservation option, i.e. those that generate triangular facets everywhere. Due to changes in the way that edges and corner facets are now shared, the pairs of triangular facets at the corners may not occur in the exact same order and may be rotated slightly. The number of coordinates and facets and the overall pattern remain the same, but it may be topologically permuted slightly at the corners.


davidgyu:
Filed as internal issue #OSD-483

PR #1387, opened 2026-08-19
Addressed warnings from Clang 21

davidgyu:


davidgyu:
Filed as internal issue #OSD-484

